PR #10, opened 2014-07-10
add credit view for credit creator

chuenlok:


mods/wikirate/sets/all/wikirate.rb
```diff
@@ -90,7 +90,20 @@ def reset_claim_counts
...
+    %{
+      <span class="credit_creator">
+        #{ action }
+        #{ _render_created_at }
+        ago by
+        #{ subformat(card.creator)._render_link }
+      </span> 
+    }
+  end
 end
```
ethn, line 107: why is action a variable here?

mods/wikirate/sets/type/webpage.rb
```diff
@@ -6,22 +6,27 @@
   linkparams = subcards["+#{ Card[:wikirate_link].name }"]
   url = linkparams && linkparams[:content] or raise "don't got it"
 
+  abort :failure, "Empty Source" if url.length == 0
```
ethn, line 9: I've previously explained the cases where we use abort(:failure).  This is not one.

If it gets this far, we should add an error.  But it really shouldn't get this far; we should check this in the interface.

mods/wikirate/sets/type/webpage.rb
```diff
@@ -6,22 +6,27 @@
...
-  subcards["+title"] = preview.title
-  subcards["+description"] = preview.description
-  subcards["+image url"] = first_image_url
+    subcards["+title"] = preview.title
+    subcards["+description"] = preview.description
+    subcards["+image url"] = first_image_url
+  rescue
+    Rails.logger.info "Fail to extract information from the #{url}"
+  end
+
```
ethn, line 29: I think the begin statement only needs to surround the LinkThumbnailer call.  (Eventually this whole thing probably needs to be broken out into cleaner methods to begin with).

If it does raise an error, do we want to add the source anyway? I kind of think not, so we should probably add an error field here, no?

mods/wikirate/spec/mods/wikirate/sets/type/webpage_spec.rb
```diff
@@ -6,8 +6,8 @@
     it "should add title,description" do
       login_as 'joe_user'	
         url = 'http://www.google.com'
-        #Card::Env.params[:sourcebox] = url
-        sourcepage = Card.create! :type_id=>Card::WebpageID,:subcards=>{ '+Link' => {'content'=> "#{url}"} }
+        Card::Env.params[:sourcebox] = true
+        sourcepage = Card.create! :type_id=>Card::WebpageID,:subcards=>{ '+Link' => {:content=> "#{url}"} }
```
ethn, line 10:
```
"#{url}"
```

is the same as just:

```
url
```

right?

mods/wikirate/sets/type/claim.rb
```diff
@@ -121,6 +121,7 @@ def analysis_names
 
 
 event :validate_claim, :before=>:approve, :on=>:save do 
+  errors.add :link, "is empty" if !subcards["+source"]||!subcards["+source"][:content]||subcards["+source"][:content].empty?
```
ethn, line 124: unreadable

mods/wikirate/sets/all/wikirate.rb
ethn, line 107: This is definitely what I suggested (thank you!), but it occurs to me that it's actually fully wagneerable.  we could do this in the claim structure:

```
{{_self | structure:creator credit}}
```

...and then in the "creator credit" card to this:

```
   added
   {{_self | created_at }}}
   ago by
   {{+*creator | core | link }}
```

What do you think? Shall we try that?

(Sorry for not thinking of that sooner)

chuenlok, line 107: Yes. I thought you want to have a new view so I followed it.
I will wagneer it.

mods/wikirate/sets/type/webpage.rb
chuenlok, line 29: I think if there is an error, we should still add the source. As the LinkThumbnailer may fail to extract information, this doesn't mean the url is invalid, maybe some bugs from LinkThumbnailer or the website added something to block people getting the meta data.

mods/wikirate/spec/mods/wikirate/sets/type/webpage_spec.rb
chuenlok, line 10: Yes. Sorry, I will change this.

mods/wikirate/sets/type/webpage.rb
chuenlok, line 9: You are right. I was thinking the abortion can save some time as the LinkThumbnailer takes some time to generate URL information, but it should return error immediately.

PR #11, opened 2014-07-10
modify based on preview pull request comments

chuenlok:
modify based on previous pull request comments
fix some rspec cases

### DIFF
--- a/mods/wikirate/sets/type/claim.rb
+++ b/mods/wikirate/sets/type/claim.rb
@@ -121,7 +121,9 @@ end
 
 
 event :validate_claim, :before=>:approve, :on=>:save do 
-  errors.add :link, "is empty" if !subcards["+source"]||!subcards["+source"][:content]||subcards["+source"][:content].empty?
+  sub_source_card = subcards["+source"]
+  #check if there is subcard +source and whether content exists and empty
+  errors.add :link, "is empty" if !sub_source_card||!sub_source_card[:content]||sub_source_card[:content].empty?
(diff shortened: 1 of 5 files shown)


mods/wikirate/sets/type/claim.rb
```diff
@@ -121,7 +121,9 @@ def analysis_names
 
 
 event :validate_claim, :before=>:approve, :on=>:save do 
-  errors.add :link, "is empty" if !subcards["+source"]||!subcards["+source"][:content]||subcards["+source"][:content].empty?
+  sub_source_card = subcards["+source"]
+  #check if there is subcard +source and whether content exists and empty
+  errors.add :link, "is empty" if !sub_source_card||!sub_source_card[:content]||sub_source_card[:content].empty?
```
ethn, line 126: much better.

sub_source_card[:content].empty? will actually cover !sub_source_card[:content], because nil.empty? is true.

so we only need the two cases.

mods/wikirate/sets/type/webpage.rb
```diff
@@ -1,33 +1,36 @@
...
 
-
 event :process_source_url, :before=>:process_subcards, :on=>:create, :when=>proc{ |c| Card::Env.params[:sourcebox] } do
-
+  
   linkparams = subcards["+#{ Card[:wikirate_link].name }"]
   url = linkparams && linkparams[:content] or raise "don't got it"
 
-  abort :failure, "Empty Source" if url.length == 0
+  errors.add :link, "is empty" if url.length == 0
```
ethn, line 8: better, thanks.

PR #13, opened 2014-07-15
create clipboard view and save related tag card when creating claim

chuenlok:


mods/wikirate/sets/right/wikirate_tag.rb
```diff
@@ -0,0 +1,9 @@
+event :create_missing_tags, :after=>:store, :on=>:save do
+	full_tagname = Card[name].content
+	tagname = full_tagname[2..full_tagname.length-3]
+	
+	if !Card.exists? tagname  
+		Card.create! :type=>"Tag", :name=>tagname,:subcards=>{'+Article' => {'content'=> '', "type_id"=>"3"} }
+	
+	end
+end
```
ethn: spacing is weird on this.  if you're using textmate, are you set to use "soft tabs" (2 spaces).

also no EOF line.

ethn: what are those subcards about? you were just trying to make it so that it creates the +Article card on the tag?

really, the only thing we want to refer to in code is the "codename". 
so it can say :type_code=>:wikirate_tag 
or :type_id=>Card::WikirateTagID

numeric references are opaque, and cardname references are fragile.

ethn: tagname = full_tagname[2..full_tagname.length-3]
??

Why not loop through the item names?  What if there are multiple new tags added?

PR #14, opened 2014-07-16
add back wikirate_tag.rb and 

chuenlok:
1. add back wikirate_tag.rb
2. add default html tooltip for the clipboard
3. change readme from text to md format
4. add config file for link_thumbnailer to repo


ethn:
Please remember to send future PR's from custom branches (not master)

script/update_tag.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    if !Card.exists? tag  
+      begin
+      Card.create! :type_id=>Card::WikirateTagID, :name=>tag
+      rescue
+        puts "invalid tag name: #{tag} FROM #{card.name}"
+      end
+    end
+  end
+end
```
ethn, line 18: EOF

mods/wikirate/sets/right/wikirate_tag.rb
```diff
@@ -0,0 +1,10 @@
+event :create_missing_tags, :after=>:store, :on=>:save do
+  
+  new_tags = self.item_names
+  new_tags.each do |tag|
+    if !Card.exists? tag  
+      Card.create! :type_id=>Card::WikirateTagID, :name=>tag
+    end
```
ethn, line 7: this is good, I'll go ahead and merge it in.

note that if you do create! (with the bang), it will raise an error if there is a problem.  if you do Card.create, it will add error messages to the card (which you can then handle).  In general, it's preferable to handle these things explicitly, but this is fine for now.

mods/wikirate/sets/right/wikirate_tag.rb
```diff
@@ -0,0 +1,10 @@
...
+  end
+  
+end
```
ethn, line 10: eof

script/update_tag.rb
```diff
@@ -0,0 +1,18 @@
...
+  
+  tag_inside = card.item_names
+  tag_inside.each do |tag|
+
+    if !Card.exists? tag  
+      begin
+      Card.create! :type_id=>Card::WikirateTagID, :name=>tag
+      rescue
+        puts "invalid tag name: #{tag} FROM #{card.name}"
+      end
```
ethn, line 15: again, this could be a little less dramatic without the bang.  Eg:

```
  c = Card.create :type_id=>Card::WikirateTagID, :name=>tag
  if c.errors.any?
    puts "invalid tag name: #{tag} FROM #{card.name}"
  end
```

just FYI.  no need to fix here.

PR #16, opened 2014-07-21
Add placeholder to wikirate link

chuenlok:
Vignesh wants to have a placeholder for the link


mods/wikirate/sets/right/wikirate_link.rb
```diff
@@ -6,7 +6,9 @@
   #site = site_card && site_card.item_names.first
   link_to "source page", card.raw_content, :target=>'source', :class=>'wikirate-source-link external-link'
 end
-
+view :editor do |args|
+  form.text_field :content, :class=>'card-content',:placeholder=>"http://example.com"
+end
```
ethn, line 11: I'll merge this, but it's not a good long term solution.  Let's not accrue a ton of custom views for placeholders, ok?

PR #18, opened 2014-07-22
Add script to add tag for too long claim

chuenlok:
Add script to add tag for too long claim


mods/wikirate/sets/type/webpage.rb
```diff
@@ -34,7 +34,7 @@ def parse_source_page url
 
 def find_duplicates url
   #need to check if content changed...
-  duplicate_wql = { :right=>Card[:wikirate_link].name, :content=>url }
+   duplicate_wql = { :right=>Card[:wikirate_link].name, :content=>url ,:left=>{:type_id=>Card::WebpageID}}
```
ethn: should not be in this branch

mods/wikirate/sets/right/wikirate_link.rb
```diff
@@ -6,7 +6,9 @@
   #site = site_card && site_card.item_names.first
   link_to "source page", card.raw_content, :target=>'source', :class=>'wikirate-source-link external-link'
 end
-
+view :editor do |args|
+  form.text_field :content, :class=>'card-content',:placeholder=>"http://example.com"
+end
```
ethn: should not be in this branch.  (branch from master first)

mods/wikirate/sets/type/wikirate_topic.rb
```diff
@@ -1,3 +1,3 @@
 view :missing do |args|
-  _render_link args
+    _render_link args
```
ethn: ?

script/add_tag_for_long_claim.rb
```diff
@@ -0,0 +1,25 @@
...
+        puts "Added tag 'name too long' : #{card.name} "
+        tags.add_item "name too long"
+        tags.save!
+      end
+    else
+      puts "Created tag 'name too long' : #{card.name} "
+      new_card = Card.create! :name=>"#{card.name}+tags",:type=>"Pointer"
+      new_card.add_item "name too long"
+      new_card.save!
+    end
```
ethn, line 14: looks good.

here's a quicker way to handle existing and non-existing cards together:

```
Card.fetch "#{card.name}+tags", :new=> {:type=>'Pointer'}
```

That will get the card if it exists and "new" it if it doesn't.

ethn, line 14: Actually, just as as exercise, want to give that a try?

PR #21, opened 2014-07-23
Check empty source when adding claim

chuenlok:
fix the rspec cases and add checking for the source when adding claim


mod/wikirate/set/type/claim.rb
```diff
@@ -81,6 +81,20 @@ def analysis_names
 
 
 event :validate_claim, :before=>:approve, :on=>:save do 
+  source_card = subcards["+source"]
+  if !source_card or !source_card[:content] or source_card[:content].empty?
+    errors.add :link, "is empty"
```
ethn: not ":link"

mod/wikirate/set/type/claim.rb
```diff
@@ -81,6 +81,20 @@ def analysis_names
...
+    else
+      duplicate_wql = { :right=>Card[:wikirate_link].name, :content=>source_card[:content] ,:left=>{:type_id=>Card::WebpageID}}
+      duplicates = Card.search duplicate_wql   
+      if !duplicates.any? 
+        errors.add :link, "is pointing to a non exisiting page" 
+      end
+    end
+  end
   errors.add :claim, "is too long (100 character maximum)" if name.length > 100
 end
```
ethn, line 113: In general, I would like to make these separate validations

mod/wikirate/set/type/claim.rb
```diff
@@ -81,6 +81,20 @@ def analysis_names
...
+    if source_card[:type_id] != Card::WebpageID
+      errors.add :link, "is pointing to invalid page" 
+    else
+      duplicate_wql = { :right=>Card[:wikirate_link].name, :content=>source_card[:content] ,:left=>{:type_id=>Card::WebpageID}}
+      duplicates = Card.search duplicate_wql   
+      if !duplicates.any? 
+        errors.add :link, "is pointing to a non exisiting page" 
+      end
+    end
+  end
```
ethn, line 112: Biggest problem:

This looks to me like it would break if I did this:

```
c = Card['myclaim']
c.name = 'the alley cat is awesome'
c.save!
```

You can confirm that by trying to rename a claim.

It's an existing claim, but I'm not adding any '+source' field; it's already there.

mod/wikirate/set/type/claim.rb
```diff
@@ -84,6 +84,42 @@ def analysis_names
...
 end
 
+event :validate_source, :after=>:approve, :on=>:save do 
+  # 1. it correctly validates when adding a claim
+  # 2. it correctly validates when editing a claim with +source
+  # 3. it doesn't break anything when editing a claim without +source (eg renaming)
+
+  #first, get the source card from request
+  source_card = subcards["+source"]||subcards["+Source"]
+
```
ethn, line 94: great inline comments!

mod/wikirate/set/type/claim.rb
```diff
@@ -84,6 +84,42 @@ def analysis_names
...
+  #second, if it is already a real card, fetch the source card
+  source_card||=fetch( :trait=>:source ) if real?
+
+  #if it is renaming, the above line wont be empty or in pointer type, maybe in search type 
+  #no need to check for this case
+  #check when null or the source_card is in pointer
+  if !real?
+    check_source source_card
+  elsif !source_card or source_card.type_id != Card::SearchTypeID
```
ethn: I'm confused.  the +Source card should always be a Pointer, right?

I still think we should eliminate the trait fetching and then just do this:

```
if source_card || new_card?
   check_source source_card
end
```

That way, if someone submits subcard data ("if source_card"), then we always check to make sure it's valid.  And if someone is creating a new card ("|| new_card?"), then we make sure it can't get created without a source.

As it stands, the code leaves the possibility that this could happen:

Person 1 goes directly to /my_claim+source and delete it.  no error.  (this is a separate problem that we need to fix later)

Person 2 goes to rename "my claim" to "your claim", and gets a weird error about the +source card.

This is really my fault, not yours, that it's so complex, but for now I think we should _only_ focus on the cases where there is an explicit subcard.  The other cases are much more involved.

We will definitely solve this in a deeper way (see http://wagn.org/implement_required_fields_on_forms), but for now let's just get rid of that trait line (96).  The logic in 103 also seems a bit off, so I'd really rather we just simplify for now.

chuenlok: I want to explain that snippet I wrote. Sorry if I made anything wrong. 

### 1. it correctly validates when adding a claim

``` ruby
source_card = subcards["+source"]||subcards["+Source"]
```

gives the source card

``` ruby
source_card||=fetch( :trait=>:source ) if real?
```

does nth as it is not real

``` ruby
if !real? or (!source_card or source_card.type_id != Card::SearchTypeID )
```

be true as it is real and go to `check_source`

### 2. it correctly validates when editing a claim with +source

``` ruby
source_card = subcards["+source"]||subcards["+Source"]
```

gives the source card

``` ruby
source_card||=fetch( :trait=>:source ) if real?
```

does nth as source_card is not null

``` ruby
if !real? or (!source_card or source_card.type_id != Card::SearchTypeID )
```

be true as source_card isnt null and type is not searchtype and go to `check_source`

### 3. it doesn't break anything when editing a claim without +source (eg renaming)

(for some cases, if users delete the content in other places(like go to +source of claim and delete all the things), and try to edit and save a claim, this could alert users as well)

``` ruby
source_card = subcards["+source"]||subcards["+Source"]
```

gives nth

``` ruby
source_card||=fetch( :trait=>:source ) if real?
```

gives the source card, but in renaming, it returns something like `<Card###2232964880[search_type:30]([card name]){new:virtual:I:nil}>`
which is in SearchType

``` ruby
if !real? or (!source_card or source_card.type_id != Card::SearchTypeID )
```

For renaming, 
be true as source_card isn't null and type is SearchType. The `check_source` is skipped.

For editing a claim without +source except renaming, 
be true as source_card isn't null and type is not SearchType and go to `check_source`

ethn: Wow, you're putting a lot of effort here, and I appreciate that.

I think the current code is fine for #1 and #2.  I think it's fragile for #3, and I would prefer to remove it.

Here is my main question: _can you explain a case where we actually do want the fetch :trait?  Is there something that breaks without it but works with it?_

As I understand things, it would only cause an error if the claim data is _already_ bad, in which case we'd be punishing the current editor for a previous editor's mistakes.

I also do not understand at all why the fetched card would be a search card.  Do you?  It seems like you're using this as a proxy to test whether we're just renaming.  That's not a good test!

chuenlok: I want to check the source before saving a claim even there is no +source included. but you are right that I should not punish the current editor for a previous editor's mistakes.

I don't know why the fetch would be a search card as well. As I wanted to check the existing source, but it gave me search card when editing the name. I have to skip the checking if the result is a search card.

PR #24, opened 2014-07-29
add popup view

chuenlok:
the parameter will be like `?show_popup=[cardname]`

### DIFF
--- a/mod/wikirate/set/all/wikirate.rb
+++ b/mod/wikirate/set/all/wikirate.rb
@@ -42,10 +42,10 @@ format :html do
     card_name = args[:show_popup]||Card::Env.params[:show_popup]
     after_card = Card[card_name]
     if !after_card
-      Rails.logger.info "Expect #{card_name} exist"
-      return ""
+       Rails.logger.info "Expect #{card_name} exist"
+    else
+       "<div class='popup-window'>#{ after_card.subformat( :format=>:html).render_core } </div>"
     end
-    content = "<div class='popup-window'>#{ after_card.format( :format=>:html).render_core } </div>"
   end
   
 =begin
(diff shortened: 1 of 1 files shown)


mod/wikirate/set/all/wikirate.rb
```diff
@@ -38,6 +38,16 @@ def cite!
...
   end
 
+  view :popup do |args|
+    card_name = args[:show_popup]||Card::Env.params[:show_popup]
+    after_card = Card[card_name]
+    if !after_card
+      Rails.logger.info "Expect #{card_name} exist"
+      return ""
+    end
+    content = "<div class='popup-window'>#{ after_card.format( :format=>:html).render_core } </div>"
```
ethn: Good stuff.  I will merge as is, but I suggest these small tweaks:
1. tiny one: you can just use params[:show_popup](you don't need Card::Env)
2. I like avoiding "return" wherever possible, so you could just use an else
3. (most important one) usually, when creating a format object within another format object, the best practice is to use "subformat", not "format", because this will then preserve a format hierarchy with parents, name contexts, etc.

so I'm picturing something like:

```
view :popup do |args|
  card_name = args[:show_popup]||Card::Env.params[:show_popup]
  after_card = Card[card_name]
  if !after_card
     Rails.logger.info "Expect #{card_name} exist"
  else
     "<div class='popup-window'>#{ after_card.subformat( :format=>:html).render_core } </div>"
  end
end
```

ethn: the rationale for avoiding "return" is that other coders can generally just go to the end of the view to see what to expect.  It's not really a deep problem, more of a stylistic thing (one which you can feel free to push back on if you care to.)

chuenlok: Thanks for the comments! I didn't know there is subformat before.

I used to have the "return" thing as it tells the reader that you don't need to read the rest which may save some time and "Exiting early allows you to pop stuff off your limited mental stack" (copy from internet to describe this :))

I will follow the current style as I don't want to have different stylistic things in the project

PR #26, opened 2014-07-30
a better way to check changes of url

chuenlok:
Sorry...I just modify it according to your comment.

### DIFF
--- a/mod/wikirate/set/right/wikirate_link.rb
+++ b/mod/wikirate/set/right/wikirate_link.rb
@@ -19,11 +19,8 @@ event :validate_content, :before=>:approve, :on=>:save do
   end
 end
 event :block_url_changing, :before=>:approve, :on=>:update do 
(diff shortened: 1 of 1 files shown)


mod/wikirate/set/right/wikirate_link.rb
```diff
@@ -19,11 +19,8 @@
   end
 end
 event :block_url_changing, :before=>:approve, :on=>:update do 
-  if real?
-    origin_card = revisions.last
-    if origin_card.content != content
-      errors.add :link, "is not allowed to be changed."
-    end
+  if real? and updates.for(:content)
+    errors.add :link, "is not allowed to be changed."
```
ethn, line 23: Thanks!

One more comment; I don't think :on=>:update will ever be called unless the card is "real".  Otherwise you're creating, not updating, right?

PR #34, opened 2014-09-04
Add function for source preview page's javascript

chuenlok:
Some functions are needed in source preview page, like checking if a website can be embed in an iframe which cannot be done in js, get source by url.


mod/wikirate/set/type/webpage.rb
```diff
@@ -9,7 +9,7 @@
   if url.length == 0
     errors.add :link, "is empty" 
   else
-    duplicates = self.type_card.find_duplicates url
+    duplicates = Card::Set::Self::Webpage.find_duplicates url
```
ethn, line 12: FYI, just Self::Webpage would probably work. (though there's nothing wrong with the full name)

mod/wikirate/set/self/webpage.rb
```diff
@@ -1,5 +1,5 @@
 
-def find_duplicates url
+def self.find_duplicates url
```
ethn, line 2: isn't it odd that this is called "find_duplicates" and not something like "find_by_url"?

ethn, line 2: or even "find_source_by_url"...

mod/wikirate/set/all/wikirate.rb
```diff
@@ -119,10 +122,52 @@ def reset_claim_counts
...
+    end
+    result.to_json
+  end
+  view :check_iframable do |args|
+    url = Card::Env.params[:url]
+    result = Hash.new 
+    counter = 0
+    result[:result]=isIframable url, counter
+    result.to_json
+  end
```
ethn, line 172: the above views seem pretty wordy.

Why

```
result = Hash.new
counter = 0
result[:result]=isIframable url, counter
```

instead of 

```
result = {:result => isIframable( url, counter=0 ) }
```

...and why

```
url = Card::Env.params[:url]
source= Card::Set::Self::Webpage.find_duplicates url
result = Hash.new 
result[:result] = false
if source.any?
   origin_page_card = source.first.left
   result[:source] = origin_page_card.name
end
```

...instead of...

```
source = Self::Webpage.find_duplicates Env.params[:url]
result = {:result => false }
result[:source] = source.first.left.name if source.any?
```

...?

ethn:
no tests for any of this?

mod/wikirate/set/self/webpage.rb
chuenlok, line 2: The function find_duplicates was for checking creating source before. I found that function can also find the source by url so I just call it directly instead of changing the name. I will change the name to a better one(the one you suggest) later.

chuenlok:
These functions are short and simple. Still need test for them?

mod/wikirate/set/all/wikirate.rb
chuenlok, line 172: I will change it.

ethn:
Yeah, we still need tests.  We've projected 99% code coverage. This doesn't mean we need separate unit tests for every little method, but we need to know the methods are getting touched by a test somewhere.

Besides, I would say "short and simple" means 1 or 2 lines.  Some of these methods are relatively involved (and should probably be broken down into short and simple ones)

PR #36, opened 2014-09-10
continuous integration testing 

xithan:
This includes a mysql dump of a wikirate test database that is used for integration testing. Don't know a good way how we can maintain this data in the future.  The problem is that `wagn seed --test-data` corrupts the wikirate db and vice versa loading the wikirate dump corrupts the test data.


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'http://rubygems.org'
 
-gem 'wagn', :path=>'/opt/wagn'
+gem 'wagn'#, :path=>'/opt/wagn'
```
ethn: I think we need this for stuff to work on johannes / kasper

config/application.rb
```diff
@@ -0,0 +1,47 @@
+require File.expand_path('../boot', __FILE__)
```
ethn: these config files are in gitignore; I don't get how they're here?

several of these would definitely break things (eg database.yml)

xithan: found the option at semaphore to add separate config files and removed them all from the PR

PR #38, opened 2014-09-19
source link is needed to be converted from '&' to '&amp;' as links store...

chuenlok:
source link is needed to be converted from '&' to '&amp;' as links stored are converted to '&amp;'. Query source page by url  will fail for this case.

There is no use to convert & to &amp; as there still a & in url which will cut the url when getting parameter url.


chuenlok:
I just found that it is only for old source
In dev.wikirate.org/Page_000000010, the link stored is with &amp; . I tried to add a source with same url again and it shows & only.

PR #43, opened 2014-09-24
handle relevant case

chuenlok:
handle relevant case and add permission control

### DIFF
--- a/mod/wikirate/set/self/wikirate_source_preview_page.rb
+++ b/mod/wikirate/set/self/wikirate_source_preview_page.rb
@@ -30,6 +30,9 @@ format :json do
     end
     return true
   end
+  view :get_user_id do |args|
+    result = {:id=>Auth.current_id}
+  end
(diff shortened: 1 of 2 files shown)


mod/wikirate/set/self/wikirate_source_preview_page.rb
```diff
@@ -48,16 +51,16 @@ def isIframable url,counter
...
     result.to_json
   end
-   view :feedback do |args|
+   view :feedback ,:perms=>:update do |args|
     url = Card::Env.params[:url]
     company = Card::Env.params[:company]
     topic = Card::Env.params[:topic]
     
-    type_of_irrelevance = Card::Env.params[:type_of_irrelevance]
+    type = Card::Env.params[:type]
```
ethn, line 59: "type" is too vague for this; it's used as a parameter in other cases.  can we go with something a little more descriptive?

PR #48, opened 2014-10-21
voting

xithan:


mod/voting/set/type/user.rb
```diff
@@ -0,0 +1,2 @@
+card_accessor :upvotes, :type=>:pointer
+card_accessor :downvotes, :type=>:pointer
```
xithan, line 2: Is user the right place or do we want this on user+*account?

btw: We have a problem with simplecov's coverage calculation here. Simplecov uses the tmp directory and because of these two lines the whole user module is  copied to  tmp/<some digits>-user.rb. But the spec for this is in wagn (or should be; currently we have no tests for user.rb) and not in wikirate so these two lines cause a dramatic decline of the the coverage result. 

two solutions: 
1) easy but poor: copy wagn rspec files to wikirate when we extend wagn's core sets
2) nice but hard: teach rspec to run automatically the necessary wagn core specs

mod/voting/spec/set/right/upvotes_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  #
+  #   it 'removes claim from upvotes pointer' do
+  #     Card::Auth.as_bot do
+  #       @uv_card.drop_item @claim.name
+  #       @uv_card.save!
+  #     end
+  #     expect(Card.fetch('joe user+*upvotes').content).not_to match "[[#{@claim.name}]]"
+  #   end
+  # end  
+end
```
xithan: So many tests and they all became useless because the logic moved from up- and downvotes to vote count :(
A last chance to do honour to them before they will be gone forever.

mod/voting/set/right/vote_count.rb
```diff
@@ -0,0 +1,192 @@
...
+    
+    .vote-details {
+    color: #bbb;
+    font-size: 14px;
+    padding-top: 7px;
+    }
+    .vote-number {
+    font-weight: bold;
+    }
+
```
xithan: this should move to *css before we deploy

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendor/wagn"]
+	path = vendor/wagn
+	url = git://github.com/wagn/wagn.git
```
xithan, line 3: Added this for continuous integration testing so that I can use wagn's github repository instead of the gem.
If we want to use this on the servers we have to change the paths in the Gemfiles

mod/voting/set/right/vote_count.rb
```diff
@@ -0,0 +1,192 @@
...
+        :class=>"slotter vote-link vote-button", :type=>'button', :title=>title, :remote=>true, :method=>'post',  :slotSelector=>".vote"}.merge(extra)) do
+      text
+    end
+  end
+
+
+  def vote_up_link
+    case card.vote_status
+    when '+'
+      disabled_vote_link :up, "You have already upvoted this claim."
```
xithan, line 97: #ENGLISH

mod/voting/set/right/downvotes.rb
```diff
@@ -0,0 +1,21 @@
...
+#   #   end
+# end
+
+def add_id new_id
+  add_item ("~#{new_id}")
+end
+
+def drop_id id
+  drop_item  ("~#{id}")
+end
```
ethn, line 7: should these be in pointer?  (in the core?)

mod/voting/set/type/user.rb
ethn, line 2: 3) make it so that you can add arbitrary names to the end of a set module file and have only that go to the tmp dir?

eg /type/user/blah.rb

In fact, there's _some_ chance I already made that work.  I know I was thinking about it...

mod/voting/spec/set/right/upvotes_spec.rb
ethn: their lives were not in vain.

mod/voting/set/right/downvotes.rb
xithan, line 7: had the same thought but was anxious that this will cause problems when we mix names and ids in pointers.

ethn, line 7: ok.  my guess is that we'll want it there eventually, but we have time to smoke out issues.

PR #56, opened 2014-10-25
fixed voting specs

xithan:
vote buttons moved from core to content view

### DIFF
--- a/mod/voting/spec/set/right/vote_count_spec.rb
+++ b/mod/voting/spec/set/right/vote_count_spec.rb
@@ -146,20 +146,20 @@ describe Card::Set::Right::VoteCount do
     end
   end
   
-  describe "core view" do
+  describe "content view" do
     before do
       Card::Auth.as_bot  do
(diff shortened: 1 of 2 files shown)


mod/voting/spec/set/right/vote_count_spec.rb
```diff
@@ -146,20 +146,20 @@
     end
   end
   
-  describe "core view" do
+  describe "content view" do
```
ethn, line 149: We should probably put together some best practices for view organization.

As a rule of thumb, I generally try not to override views like content, open, titled, closed, etc.  My idea has been that those should all have the same core content ("core"), and that any other view of the core content (eg the voting interface) should have a different name.

But there are some patterns emerging that challenge this, especially cases where titled and open need different headers.  Maybe we should think about API for that...

ethn, line 149: (nothing urgent)

PR #57, opened 2014-10-27
show clipboard in the front if not cited

chuenlok:


mod/wikirate/set/type/claim.rb
```diff
@@ -135,9 +135,8 @@ def check_source source_card
 
 view :title do |args|
   %{ 
-    #{ args[:citation_number] }
+    #{ args[:citation_number]||(optional_render :clipboard, args, :hide) }
```
ethn: if we want these to show up to the left of the voting buttons, neither the citation number nor the clipboard can be inside the title, right?

chuenlok:
I will modify the CSS for this after deployed.

PR #59, opened 2014-10-28
colors for citation boxes

xithan:


mod/wikirate/set/right/citation_count.rb
```diff
@@ -0,0 +1,6 @@
+format :html do 
+  view :titled do |args|
+    args.merge!(:slot_class=>( card.format.render_core=="0" ? "red-citation" : "green-citation" ))
```
ethn, line 3: annoying nitpick: css class should describe function, not form (eg, in case we change colors).

(Sometimes it's nice I'm on a separate continent; it means I get slapped less.)

mod/wikirate/set/type/claim.rb
```diff
@@ -68,12 +68,24 @@ def next_step_tip
   end
   
   view :header do |args|
-    if args[:home_view] == :open and !args[:without_voting] 
-      render_header_with_voting
-    else
-      super(args)
-    end
-  end
+     if args[:home_view] == :open
```
ethn, line 71: I think this – args[:home_view] – is why the header is disappearing upon reload.  the home_view arg is generated by #nest, but in the context of an AJAX call, #nest is not called.

ethn, line 71: we really need to be able to traverse a view tree, don't we?

ethn, line 71: maybe something like this:

```
view :open do |args|
  super args.merge( :custom_claim_header=>true )
end
```

...then you can just look for that arg?

mod/wikirate/set/right/citation_count.rb
xithan, line 3: the function of that class is to make the citation red ;) 
I'll rename it "no-citations"

ethn, line 3:
```
<div class="to-make-ethan-shut-up"></div>
```

mod/wikirate/set/type/claim.rb
xithan, line 71: that's a good solution

PR #62, opened 2014-10-29
vote and source and claim structure

chuenlok:
I cannot get the vote header for claim correctly. I think I need some scripts to migrate the vote count stuff. 
I tested the the citations box and claim box and seems working normally. I will test them all again tmr


chuenlok:
Tested and it seems working normally.

PR #63, opened 2014-10-30
basic claim filter working

ethn:
This one's not quite ready either, but the form is now working.  We still need to make the filter use a real card for paging and such, to get the filter box working with autocomplete, and to clean up the whole thing.


xithan:
merged in my claim filter PR

PR #65, opened 2014-10-30
Related article

chuenlok:
show the editor of article in analysis page
show claim tip when there is no related article


mod/wikirate/set/right/related_articles.rb
```diff
@@ -12,7 +12,9 @@
             #{ next_action_link analysis_name.to_name}
           </div>
         }
-      end.join ' ')
+      end.join ' '
+      articles = "<span class='no-article'>No related Article</span>"+ claim.format.render_tips if analysis_names.length == 0
+      process_content( articles )
```
ethn, line 17: any idea why "process_content" is needed here?

mod/wikirate/set/right/related_articles.rb
```diff
@@ -23,8 +25,10 @@ def next_action_link analysis_name
       when !article.includees.include?( card.left); 'Cite'
       else                                          'View'
       end
-    opts = { :edit_article=>true }
-    opts[ :citable ] = card.cardname.trunk_name unless act =~ /^Edit/
+    if act == "Cite"
+      opts = { :edit_article=>true }
+      opts[ :citable ] = card.cardname.trunk_name 
+    end
     %{ <span class="claim-next-action">[[/#{analysis_name.url_key}?#{opts.to_param} | #{ act }]]</span> }
```
ethn, line 32: do we want this at all if the act isn't "Cite"?

PR #81, opened 2014-11-18
Source preview optimization

chuenlok:


chuenlok:
I don't know what the fail case fails. It is working normally on my local machine and I try to simulate the request manually on dev and live with correct results.

ethn:
How close is this?

chuenlok:
I removed some duplicated codes to functions. Waiting for your review...

PR #82, opened 2014-11-20
fix duplicated sources adding

chuenlok:


ethn:
I didn't study this in great detail, but it looks from a quick glance like well organized and tested code, so I'm going to push it live.

PR #92, opened 2014-12-03
fix some redirection problem

chuenlok:
short url (http://wh.gov/iO3vF) will cause the redirection error.


mod/wikirate/set/self/wikirate_source_preview_page.rb
```diff
@@ -167,7 +167,7 @@ def is_iframable? url, request
     begin 
       # escape space in url, eg, http://www.businessweek.com/articles/2014-10-30/tim-cook-im-proud-to-be-gay#r=most popular
       url.gsub!(/ /, '%20')
-      uri = open(url)
+      uri = open(url, :allow_redirections => :safe)
```
ethn, line 170: is this because CERTH is still sending us urls that need redirection?

chuenlok, line 170: Even CERTH does not send us the short URL, some long url will also have the redirection like www.google.com (but google bans the iframe).

PR #94, opened 2014-12-12
update according to certh api

chuenlok:
add company and topic names when sending feedback


chuenlok:
Did we exclude all files in config folder? The initializer of link_thumbnailer is added back suddenly

xithan:
The initializer of link_thumbnailer moved to the wikirate mod. Please remove it from you config folder and push again.

PR #105, opened 2015-02-02
Wagny source preview

chuenlok:
I am still working on the js. I will finish it tmr and write tests for this commit.


ethn:
How are there 677 additions and only 9 deletions?  Isn't a lot of this just moved code?

ethn:
If there are files to delete, it will help to delete them; I think git may recognize it as a renaming (though you may need to rebase?)

chuenlok:
I did not remove the old source preview page code. I will remove it today later

ethn:
there is no migration here.  Won't we need to update lots of wagneering (links and such) to get all this to work?

chuenlok:
I thought doing some wagneering by hands. I will make a migration for this tmr

ethn:
I was assuming you'd do the wagneering by hand and then export.  Is that what you mean?

chuenlok:
I thought update the wagneering stuff on dev by hands. Nevermind, I made a migration for this. :)

ethn:
Wagneering on dev is a totally acceptable process, but we would need to be organized with that.  This puts our master code in a "ready for dev but not ready for live" state, and we need to be able to see that well.

How should we flag that?  Perhaps with a github label (https://help.github.com/articles/creating-and-editing-labels-for-issues-and-pull-requests/)?  Or perhaps we should use a different branch on dev than on live?

Now that we're using submodules, the different branch idea might make sense...

PR #106, opened 2015-02-10
overwrite the google analytic js to add content group

chuenlok:


chuenlok:
I tried to overwrite the raw view and call the `super args` but it fails so i overwrite the whole function directly.  Please let me know if it is not the correct way.

chuenlok:
I am thinking whether I should do it in this way or not inside the `mod/wikirate/set/self/head.rb`

``` ruby
    original_js = super
    %{
      <script type="text/javascript">
        var _gaq = _gaq || [];
        _gaq.push(['_setPageGroup', '1', '#{root.card.type_name}']);
      </script>
      #{original_js}
    }
```

ethn:
I went ahead and merged what you'd done, but I do like the solution overriding less code.

We probably want the API to be quite different, so feel free to propose changes to how Wagn does this.  Ultimately we don't want Google Analytics hard coded in such a core place in Wagn, but I haven't really put together a better proposal for how it should look.

PR #108, opened 2015-02-16
Wikirate - Cite button interaction on article page 

srivig:
@chuenlok  Thanks.


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: ./vendor/wagn
   specs:
-    wagn (1.14.8)
+    wagn (1.14.7)
```
ethn, line 4: looks like you're moving to an older version :)

I'll fix that, but to get the latest try `git submodule update --remote`

PR #116, opened 2015-03-09
add layout to page link in source preview

chuenlok:


mod/wikirate/set/type/webpage/source_preview.rb
```diff
@@ -169,7 +169,7 @@ def show_options source_from_certh,source_page_name,url
 
       result = %{
         <div id="source-page-link" class="mark-irrelevant-button" >
-          <a href="/#{source_page_name}" id="source-page-button" target="_blank">
+          <a href="/#{source_page_name}?layout=wikirate layout" id="source-page-button" target="_blank">
```
ethn, line 172: underscores look cleaner than spaces in the result url (after clicking), because usually spaces get turned into %20.  But this is good; merge at will!

PR #119, opened 2015-03-16
Fix spec

chuenlok:
This replaced some exact html matching code with better html element checking.

### DIFF
--- a/lib/tasks/wikirate.rake
+++ b/lib/tasks/wikirate.rake
@@ -78,6 +78,17 @@ namespace :wikirate do
           card_to_be_kept+= "#{c.id},"
         end
 
+        type_ids = ""
+        type_ids_without_company_and_topic = ""
+        %w{ Claim Company Market Issue Topic Analysis Task Newspaper Book Activists Donor Website Person Institution Donor Status Organization Periodical Page  }.map do |typename|
+          if id = Card.fetch_id(typename) 
+            type_ids+="'#{id}',"
+            if typename!="Company" and typename!="Topic"
+              type_ids_without_company_and_topic+="'#{id}',"
+            end
+          end
+        end
+
         ActiveRecord::Base.connection.execute 'delete from card_revisions'
         #ActiveRecord::Base.connection.execute 'delete from card_actions'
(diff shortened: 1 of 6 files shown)


lib/tasks/wikirate.rake
```diff
@@ -89,13 +100,13 @@ namespace :wikirate do
...
+        ActiveRecord::Base.connection.execute "delete from cards where type_id = '#{Card::WikirateTopicID}' and id not in ( #{topic_ids[0...-1]} )"
 
         # delete all webpage++link
-        ActiveRecord::Base.connection.execute "delete ca from cards ca inner join cards le ON ca.left_id = le.id where le.type_id in ('631','651','699','974','1010','1109','1591','1638','1690','2207','2317','2327','2754','2755','2813','2995','4010','4030') and ca.id not in  ( #{card_to_be_kept[0...-1]} )" 
-        ActiveRecord::Base.connection.execute "delete from cards where type_id in ('631','699','974','1109','1591','1638','1690','2207','2317','2327','2754','2755','2813','2995','4010','4030') and id not in ( #{card_to_be_kept[0...-1]} )"
+        ActiveRecord::Base.connection.execute "delete ca from cards ca inner join cards le ON ca.left_id = le.id where le.type_id in (#{type_ids[0...-1]}) and ca.id not in  ( #{card_to_be_kept[0...-1]} )" 
+        ActiveRecord::Base.connection.execute "delete from cards where type_id in (#{type_ids_without_company_and_topic[0...-1]}) and id not in ( #{card_to_be_kept[0...-1]} )"
```
ethn: Ahh, MUCH more readable!

ethn:
good stuff.  will merge in.  very close now!

lib/tasks/wikirate.rake
xithan: I guess the `[0..-1]` is there to get rid of the comma at the end? -1 includes the end of the string so it should say `[0..-2]` or maybe easier to read: collect the ids in an array and use `join ','`. Then you get rid of the initialization statement as well `type_ids = %w{ ... }.map { ... }.join ','`

chuenlok: let me try your comment :+1:

PR #124, opened 2015-03-19
Reorganize source

chuenlok:


mod/wikirate/set_pattern/citable.rb
```diff
@@ -0,0 +1,17 @@
...
+  { :name=>"*dummy", :type_id => Card::SourceFileID }
+end
+
+def pattern_applies? card
+  card.type_id == Card::SourceFileID || card.type_id == Card::WebpageID || card.type_id == Card::TextSourceID 
+end
+
+def follow_label name
+  label name
+end
```
xithan, line 17: this should return something shorter so that it fits in the gear menu. I suggest "all citeable sources". It's also important that the first letter is lowercase because in the follow submenu the word "follow" is prepended

db/migrate_cards/20150318162100_import_new_source_type_and_set_pattern.rb
```diff
@@ -0,0 +1,16 @@
...
+class ImportNewSourceTypeAndSetPattern < Card::Migration
+  def up
+    Card.create! :name=>"TextSource", :type_id=>Card::CardtypeID, :codename=>"text_source"
+    
+    Card.create! :name=>'Citable', :type_id=>Card::SetID, :codename=>'citable' 
+      ,:content=>%{
+        {
+          "type":["in","SourceFile","Page","TextSource"]
+        }
+      }
```
xithan, line 11: Maybe it's time to rename 'Page' to 'WebSource'?  I think it should be consistent how we compose the names of those types. Either SourceFile and SourceText or FileSource and TextSource.

chuenlok, line 11: I will do the renaming and re-organize the source part as a mod later.

ethn:
Note that if/when we rename "page" to "Web Source", this will affect a lot of WQL and possibly some CSS (neither of which are automatically updated)

ethn:
I still think we should do it, but please expect breakage.  And plan fixage!

PR #125, opened 2015-03-19
fix stack level too deep problem for claim's tip view

chuenlok:
claim structure card has`{_self|tip}`. tip view is not rendered in closed view so it will render the closed view again.

### DIFF
--- a/mod/wikirate/set/type/claim.rb
+++ b/mod/wikirate/set/type/claim.rb
@@ -73,7 +73,7 @@ format :html do
   end
 
   
-  view :tip, :perms=>:none do |args|
+  view :tip, :perms=>:none, :closed=>:blank do |args|
     # special view for prompting users with next steps
     if Auth.signed_in? and ( tip = args[:tip] || next_step_tip ) and @mode != :closed
       %{


ethn:
I merged this into the ratings branch but haven't done master yet.  Any idea why this is failing even though master is not?

xithan:
outdated submodules ...

ethn:
OH.  This is a flaw in the test setup, imo.

It runs `git submodule update --remote`, which updates to HEAD of wagn/wagn master.  If we just run `git submodule update`, it will use the SHA that we store in the repo.

I'll make that change but won't re-run these tests just yet.  (would rather use the test time on the current stuff)

vendor/wagn
```diff
@@ -1 +1 @@
-Subproject commit bee8b896f05495d9eb741b048bedc663c911ba3e
+Subproject commit 03cb9f7cc83831c3c0b69e5032fcbe17067a7e28
```
ethn: I think we probably actually want to bring this back to the commit that was in the original PR.  I updated the test on semaphore NOT to use the latest wagn code but respect the commit indicted in the submodule.

The latest wagn code is a bit ahead of wikirate master now.

ethn: in other words, we want it so that the _only_ change to master is claim.rb.  I know that didn't work before, but I think it will now, thanks to the semaphore change.

PR #133, opened 2015-03-23
add view source button in metric details

chuenlok:


mod/rating/set/right/values.rb
```diff
@@ -10,6 +10,7 @@ def timeline_row value_card, connect=false
         content_tag(:div, year.html_safe, :class=>'year td'), 
         content_tag(:div, value, :class=>'value td' ),
         content_tag(:div, credit.html_safe, :class=>'credit td'),
+        (link_to 'View Source(s)', "/#{card.left.name}+#{value_card.cardname.right}+Summary",:class=>'btn btn-yellow btn-xs view-source-btn'),
```
xithan: two suggestions:
- use `card_link  <cardname>, :text=>..., :class=>...` instead of `link_to`
- use ~~`card.cardname.left`~~ even better: `card.trunk` instead of `card.left.name`.  card.left fetches the left card, card.cardname.left only operates with strings

PR #140, opened 2015-03-25
update cap script

chuenlok:
update `get_backup` to `backup:get` and move to namesapce `backup`, same case as pull_db


ethn:
cool. this is better.

PR #153, opened 2015-04-17
source box handles card name

chuenlok:


mod/wikirate/set/right/source.rb
```diff
@@ -5,7 +5,7 @@
         <div class="source-editor nodblclick">
           #{ form.hidden_field :content, :class=>'card-content' }
           <div class="sourcebox">
-            #{ text_field_tag :sourcebox, nil, :placeholder=>'http://' }
+            #{ text_field_tag :sourcebox, nil, :placeholder=>'card name or http://' }
```
ethn, line 8: Wikirate users don't know what a "card" is.  Maybe "source"?

PR #164, opened 2015-04-30
handle multiple topic in csv

chuenlok:
Sorry. Just found that Lucy had put more than 1 topic in cells.


db/migrate_cards/20150429152546_import_new_metric.rb
```diff
@@ -8,7 +8,13 @@ def up
...
+        if topics = row[:topics]
+          topic_str = ""
+          topics.split(";").each do |t|
+            topic_str+="[[#{t}]]\n"
+          end
+          subcards.merge!({'+topics'=>{:content=>topic_str,:type_id=>Card::PointerID}}) 
+        end
         subcards.merge!({'+question'=>{:content=>row[:question]}}) if row[:question]
         subcards.merge!({'+unit'=>{:content=>row[:unit].to_s,:type_id=>Card::PhraseID}}) if row[:unit]
         subcards.merge!({'+range'=>{:content=>row[:range].to_s,:type_id=>Card::PhraseID}}) if row[:range]
```
ethn, line 20: Why are there metrics in CSVs?  In the future all metrics should be added by hand, correct?

ethn, line 20: You can merge this and run it, but I don't want this to be a common pattern.

ethn, line 20: Meaning I want folks to create Metrics on the site and use/test the interface for it.

PR #169, opened 2015-05-08
Import metric

chuenlok:
new metrics from Richard and Lucy


vendor/wagn
```diff
@@ -1 +1 @@
-Subproject commit dc80edaa632dcfc866ab1d1121f599cba608d230
+Subproject commit fcf1e61a9a0e818c44ca5b58dfdcd12a429b6665
```
chuenlok: I don't know why it said  `vendor/wagn` is changed as the current submodule version in `wagn/wikirate` is `fcf1e61a9a0e818c44ca5b58dfdcd12a429b6665` already

PR #174, opened 2015-05-13
Fix metric value adding

chuenlok:


mod/rating/set/right/add_value.rb
```diff
@@ -2,14 +2,23 @@ def virtual?; true end
...
+          :type_id=>Card::MetricValueID,
+          :subcards=>{
+            '+value'=>{:content=>Env.params[:value]},
+            '+source'=>"[[#{source_card.name}]]"
+          }}}
+    else
+      source_card.errors.each do |key,value|
+        errors.add key,value
+      end
+      abort :failure
```
chuenlok, line 20: If I don't abort failure, it will continue and cause other weird exceptions.

ethn, line 20: this API needs serious work!

PR #177, opened 2015-05-13
add type for existing aliases card

chuenlok:


script/import_companies.rb
```diff
@@ -18,7 +18,7 @@
       if wikirate_company_name = row[:wikirate_company]
         aliases_card = Card.fetch("#{wikirate_company_name}+aliases") || Card.create!(:name=>"#{wikirate_company_name}+aliases",:type_id=>Card::PointerID)
         aliases_card.content = "#{alias_content}"
-        
+        aliases_card.type_id = Card::PointerID
```
ethn, line 21: fine to have here, but there should probably also be a card with "aliases+_right+_default"

PR #178, opened 2015-05-16
Metric polish

chuenlok:
to show the latest metric value on the metric, company row


mod/rating/set/right/add_value.rb
```diff
@@ -1,18 +1,32 @@
 def virtual?; true end
 
 event :add_value, :before=>:approve, :on=>:update do
+  add_metric_value
+end
+
+event :add_value, :before=>:approve, :on=>:create do
+  add_metric_value
+end
+
```
ethn, line 10: if it's supposed to happen on update, you can just use :on=>:save. 

that would probably also warrant a comment as to why we're doing an "add" on an update.  I think I get it, but we need more comments!

PR #179, opened 2015-05-18
add missing handling on company+about

chuenlok:


mod/rating/set/type_plus_right/metric/add_value.rb
```diff
@@ -8,28 +8,32 @@ def value_name
...
-          wrap_with(:div, :class=>'row') do [
-            wrap_with(:div, "<label>Year</label>#{year_tag}".html_safe, :class=>'metric-value-year col-xs-2'),
-            wrap_with(:div, "<label>Value</label>#{value_tag}".html_safe, :class=>'metric-value-value col-xs-10'),
-            ]
-          end
-        ]
-      end.concat(_render_add_source(args))
+    if not Auth.signed_in? 
+      card.format.render_denial :hide=>"header menu"
+    else
```
ethn: usually we don't want permission handling inside the view.  why is this needed?

chuenlok: I tried to add read permission for `metric+add value` but the modal fails to load the content when the http status is not 200. I will try to find a way to fix it tmr.

chuenlok:
removed the sign in checking in code.

ethn:
this is good. let's merge and deploy it.

PR #180, opened 2015-05-20
Merge source file to source

chuenlok:


mod/wikirate_source/spec/set/type_plus_right/source/file-import_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    end
+  end
+  describe "while rendering import view" do
+    
+    it "shows field correctly" do
+      
+      source_file_card = @source.fetch :trait=>:file
+      html = source_file_card.format.render_import      
+
```
ethn, line 50: I think I would be tempted to do binding.pry, run this both individually and with all specs, and then see what you can do to find something different in the two.  check Card::Env, Card::Auth, the local cache, etc.  Just be careful not to change much in the whole-spec environment (so hard to get back there again!)

chuenlok, line 50: I will restart the PR after I found it.

PR #181, opened 2015-05-20
Fix claim source reference

chuenlok:


script/fix_claim_reference.rb
```diff
@@ -0,0 +1,16 @@
...
+
+Card::Auth.as_bot do
+
+  claim_cards = Card.search :type_id=>Card::ClaimID,:not=>{:right_plus=>[Card[:source].name,{:refer_to=>{:type_id=>Card::SourceID} }]}
+  claim_cards.each do |card|
+    claim_source = card.fetch :trait=>:source
+    if claim_source
+      puts claim_source.name
+      claim_source.update_references
+      claim_source.expire_structuree_references
```
ethn: I think expire_structuree_references only applies to cards that are structures (in fact, that method and several others in templating.rb should probably be in right/structure.rb).  So I doubt this line does everything.

script/fix_claim_reference.rb
```diff
@@ -0,0 +1,16 @@
+require File.expand_path('../../config/environment',  __FILE__)
+
+
+Card::Auth.as_bot do
+
+  claim_cards = Card.search :type_id=>Card::ClaimID,:not=>{:right_plus=>[Card[:source].name,{:refer_to=>{:type_id=>Card::SourceID} }]}
+  claim_cards.each do |card|
+    claim_source = card.fetch :trait=>:source
+    if claim_source
+      puts claim_source.name
```
ethn: the above is fine, but I think you could also do 

```
Card.search :right=>Card[:source].name, :not=>{:refer_to=>....
```

then you would have all the +source cards.

Doesn't matter really, just thought you might find it interesting.

ethn: actually, I guess it matters just a tiny bit, in that we might need to reuse this in the future if the reference problem recurs.  If you do a little clean up, I think it's ready to merge and run.

ethn:
looks good to go.

PR #184, opened 2015-05-28
Merge source file to source

chuenlok:


mod/wikirate/set/type/file.rb
```diff
@@ -4,7 +4,9 @@
     if args[:structure]
       process_content _render_raw(args)
     else
-      super args
+      handle_source args do |source|
+        card_url source
+      end
```
ethn: this seems like a significant change.  what's it about?

mod/wikirate_source/set/type/source/source.rb
```diff
@@ -36,7 +36,7 @@ def indirect_contributor_search_args
 
 event :process_source_url, :before=>:process_subcards, :on=>:create do
 #, :when=>proc{    |c| Card::Env.params[:sourcebox] == 'true'  } do
-  
+  return if subcards["+File"] or subcards["+Text"]
```
ethn: if you don't want the event to run, then the test should be in a :when arg

mod/wikirate_source/set/type_plus_right/source/file/import.rb
```diff
@@ -19,6 +21,8 @@
 
 
 event :import_csv, :after=>:store, :on=>:update do
+
+  return if Env.params["is_metric_import_update"] != 'true'
```
ethn: this is another :when

mod/wikirate_source/set/type_plus_right/source/file/import.rb
```diff
@@ -108,6 +115,10 @@ def default_import_args args
     nest card.left.metric_card, :view=>:edit_in_form
   end
 
+  view :hidden_field do |args|
+    hidden_field_tag :is_metric_import_update, 'true'
+  end
```
ethn, line 116: this is a very generic name for such a specific view.  I get that it's a narrow set, but it could still be confusing.

mod/wikirate_source/spec/set/type_plus_right/source/import_test.csv
```diff
@@ -0,0 +1,4 @@
+Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live,43
+Amazon.com,9
```
ethn, line 2: :)

mod/wikirate_source/spec/set/type_plus_right/source/file-import_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      source_file.update_attributes :subcards=>{"#{@source.name}+#{Card[:metric].name}"=>{:content=>'[[Access to Nutrition Index+Marketing Score]]',:type_id=>Card::PointerID}}
+
+      
+      expect(source_file.errors).to have_key(:content)
+      expect(source_file.errors[:content]).to include("Please give a year.")
+      
+      source_file.update_attributes :subcards=>{"#{@source.name}+#{Card[:year].name}"=>{:content=>'[[2015]]',:type_id=>Card::PointerID}}
+      
+      expect(source_file.errors).to have_key(:content)
+      expect(source_file.errors[:content]).to include("Please give a metric.")
```
ethn, line 23: yay for this!

ethn:
please address minor comments and merge.

mod/wikirate/set/type/file.rb
chuenlok: I tried to call `super args` to show the default core view when there is no structure. I found that the `view :core` in `mod/wikirate/set/type/file.rb` will rewrite the `view :core` in `file.rb` in wagn and the `super args` will call the upper layer's core view from `format do` instead of `format :html do` in file.rb . I found a new way to overcome this problem by having one more submodule. `super` is back

ethn: Your new solution is exactly right.

mod/wikirate_source/set/type/source/source.rb
ethn: Thanks for fixing this.  I should explain the rationale here.  If you put the test in :when, then the event does not actually happen when the test fails.  This means that any event that is supposed to happen before ore after process_source_url will not get triggered.

PR #185, opened 2015-05-28
progress bar

chuenlok:


db/migrate_cards/data/progress_bar.json
```diff
@@ -0,0 +1,42 @@
...
+  "timestamp": "2015-05-28 21:03:51 +0100",
+  "card": {
+    "name": "export",
+    "type": "Pointer",
+    "content": "[[progress bar]]\n[[Project]]\n[[progress bar+*right+*structure]]\n[[project task progress]]\n[[campaign+*type+*structure]]\n[[project with progress bar]]",
+    "value": [
+      {
+        "name": "progress bar",
+        "type": "Cardtype",
+        "codename": "progress_bar"
```
ethn: looks like this type was never used?

db/migrate_cards/data/progress_bar.json
```diff
@@ -0,0 +1,42 @@
...
+      },
+      {
+        "name": "Project",
+        "type": "Cardtype",
+        "codename": "project"
+      },
+      {
+        "name": "progress bar+*right+*structure",
+        "type": "HTML",
+        "content": "<div class=\"progress\">\r\n  <div class=\"progress-bar\" role=\"progressbar\" aria-valuenow=\"{{_l|raw}}\" aria-valuemin=\"0\" aria-valuemax=\"100\" style=\"width: {{_l|raw}}%;\">\r\n    {{_l|raw}}%\r\n  </div>\r\n</div>"
```
ethn: any reason not to associate this with the type?

mod/wikirate/set/type_plus_right/project/project_task_progress.rb
```diff
@@ -0,0 +1,15 @@
...
+    
+    closed_task_number = Card.search :found_by=>card.cardname.left+"+tasks",:right_plus=>["Status",{:refer_to=>"Closed"}],:return=>"count"
+    all_task_number = Card.search :found_by=>card.cardname.left+"+tasks",:return=>"count"
+
+    if all_task_number == 0 
+      0
+    else
+      (closed_task_number.to_i*100)/all_task_number.to_i
+    end
+  end
```
ethn: I kind of think we should leave this out for now and just make it editable.  Not sure the tasks will be as good of an indication of progress as the metric values in many cases...

db/migrate_cards/data/progress_bar.json
chuenlok: I am thinking that if the right structure can show the progress bar based on the left value then we don't need to create a type. I think it is a minimum way to show a progress bar?

mod/wikirate/set/type_plus_right/project/project_task_progress.rb
chuenlok: We can have `project+progress` as a number. `project+progress+progress bar` can show the progress bar based on the number users input in `project+progress` card.

I can remove this code if the above solution is fine to you.

db/migrate_cards/data/progress_bar.json
chuenlok: It should be a basic type. I will change it.

db/migrate_cards/data/progress_bar.json
ethn: You're probably right that this is minimal, but it's not super flexible.  If, for example, I wanted to create:
- MyProject+progress on metrics
- MyProject+progress on tasks

I couldn't use that naming pattern.  And in general, I'm stuck with using your name. 

If we go with the type, we can probably stick this in a mod and make it usable in many other contexts.

Also, you probably need to change it so that the editor is like a Phrase (not Basic) editor so that the percentage value doesn't get wrapped in paragraph tags.

mod/wikirate/set/type_plus_right/project/project_task_progress.rb
ethn: now that I think about it, why don't we just make "progress_bar" a view?  That would be even more flexible.  So then you could do {{+progress|progress_bar}}.

I think this makes most sense.

db/migrate_cards/data/progress_bar.json
ethn: (skip the above, I think the solution below is better)

mod/wikirate/set/all/wikirate.rb
```diff
@@ -19,6 +19,25 @@
...
+    if is_number? value
+      %{
+        <div class="progress">
+          <div class="progress-bar" role="progressbar" aria-valuenow="#{value}" aria-valuemin="0" aria-valuemax="100" style="width: #{value}%;">
+            #{value}%
+          </div>
+        </div>
+      }
+    else
+      "Only card with numberic content can be shown as progress bar."
```
ethn: no "b" in "numeric"

mod/wikirate/set/all/wikirate.rb
```diff
@@ -19,6 +19,25 @@
     CGI::escape card.name
   end
 
+  def is_number? str
+    true if Float(str) rescue false
+  end
+
+  view :progress_bar do |args|
+    value = card.db_content
```
ethn: This should be "raw_content".

Definitions:
card.db_content = value in database
card.content = db_content unless new? (in which case get from default rule)
card.raw_content = content unless structured (in which case get from structure rule)

Arguably, those names could us some improvement.  But in this case if we do "db_content", then we'll never be able to use progress bars for dynamic (structured) calculations.

mod/wikirate/set/all/wikirate.rb
```diff
@@ -19,6 +19,25 @@
     CGI::escape card.name
   end
 
+  def is_number? str
+    true if Float(str) rescue false
+  end
```
ethn, line 24: this is good for now.  Feels like this should be a method on a content object, but we don't have a fantastic framework for that yet.

ethn:
two small suggestions made.  we can merge and deploy after that.

mod/wikirate/set/all/wikirate.rb
chuenlok, line 24: or we can put it to String class?
Like:

```
class String
  def is_number?
    true if Float(self) rescue false
  end
end
```

PR #186, opened 2015-05-28
fix while adding metric value in metric progress page

chuenlok:
I don't know there is another way to add metric value apart from `metric+add value`. I will start writing test cases for the rest of wikirate code. It is too tired to find bugs one by one by users.

### DIFF
--- a/mod/rating/set/right/add_value.rb
+++ b/mod/rating/set/right/add_value.rb
@@ -8,6 +8,13 @@ event :add_value, :before=>:approve, :on=>:create do
   add_metric_value
 end
 
+def value_name
+  if (metric_and_company_name = cardname.left) && Env.params[:year] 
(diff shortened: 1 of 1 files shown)


mod/rating/set/right/add_value.rb
```diff
@@ -8,6 +8,13 @@ def virtual?; true end
   add_metric_value
 end
 
+def value_name
+  if (metric_and_company_name = cardname.left) && Env.params[:year] 
+    "#{metric_and_company_name}+#{Env.params[:year]}"
+  end
+end
+
```
ethn, line 16: this looks fine to merge!

ethn:
As a rule, we should start by fixing bugfixes like this by reproducing the bug in a test.  I know we haven't been strict about that before, but I think we need to start that now.

ethn:
...or tomorrow.  I'm going to merge this :)

PR #192, opened 2015-06-04
Source preview polish

chuenlok:
hide the menu and toolbar in some popup window and updated the rspec


ethn:
merged this, but in general I think we need the menu more, not less!

PR #193, opened 2015-06-08
facebook meta tags

chuenlok:


mod/wikirate/set/type/image/wikirate_image.rb
```diff
@@ -0,0 +1,8 @@
+view :missing do |args|
+  if home_view = args[:home_view] and home_view == :source
+    # to get the wikirate icon
+    Card["*logo"].format.render_source args.merge({:size=>"large"})
```
chuenlok: I am not sure whether this approach is fine or not. 

I need to get the wikirate logo url while that image is missing for the facebook meta

db/migrate_cards/data/facebook_meta.json
```diff
@@ -0,0 +1,106 @@
...
+    "value": [
+      {
+        "name": "facebook meta+*right+*structure",
+        "type": "HTML",
+        "content": "<div style=\"display:none\">\r\n  <meta property=\"og:site_name\"          content=\"WikiRate.org\" />\r\n  <meta property=\"og:url\"                content=\"http://wikirate.org/{{_l|linkname}}\" />\r\n  <meta property=\"og:type\"               content=\"{{_l+facebook_meta_type|core}}\" />\r\n  <meta property=\"og:title\"              content=\"{{_l+facebook_meta_title|core}}\" />\r\n  <meta property=\"og:description\"        content=\"{{_l+facebook_meta_description|core}}\" />\r\n  <meta property=\"og:image\"              content=\"http://wikirate.org{{_l+facebook_meta_image|core}}\" />\r\n</div>"
+      },
```
ethn, line 19: As I understand it, these meta tags only really matter when facebook itself is making the request.  Therefore, I was thinking perhaps we should only go through the process of rendering all these cards when we know that facebook is the requester.

We could still use the above cards, but we would probably need to add code to identify the facebook agent.

mod/wikirate/set/type/image/wikirate_image.rb
ethn: I never love adding a lot of logic to the core views to handle special cases.

ethn: perhaps it makes more sense to create a view like "og_source"?

mod/wikirate/set/self/head/wikirate_head.rb
```diff
@@ -16,6 +16,15 @@ def google_analytics_head_javascript
       }
     end
   end
+
+  view :raw do |args|
+    result = super(args)
+    user_agent = request.user_agent
+    if user_agent=="Facebot" or user_agent.include?"facebookexternalhit/1.1"
+      result += subformat(Card.fetch("#{Env.params["id"]}+facebook_meta"))._render_core
+    end
```
chuenlok, line 25: I think I can only handle this special case in this view?

chuenlok:
I encounter a problem for homepage's description. 

I am using `facebook_meta+*right+*structure` and  `facebook_meta_description+*right+*structure`. In the homepage, I cannot specify a set for `facebook_meta_description` structure. It can only belongs to 
###
All "+facebook meta description" cards on "HTML" cards
All "+facebook meta description" cards (current)
All "HTML"+"Basic" cards
All "Phrase" cards
All "+" cards
All cards
###
Can I specify a structure rule for specific card `home+facebook_meta_description`?

ethn:
yes, you can specify that rule, but you'll probably have to create the actual card home+facebook_meta_description first.

mod/wikirate/set/self/head/wikirate_head.rb
ethn, line 25: sounds good.

ethn:
failing on semaphore?

chuenlok:
Something the link_thumbnailer fails to get link meta data on semaphore server. I tried to reproduce the problem locally or the command prompt provided by semaphore to run test on their server but no luck. Rebuild the test on semaphore can solve the problem mostly.

PR #198, opened 2015-06-12
show empty contribution cards in closed view

xithan:


mod/wikirate/set/right/contributed_analysis.rb
```diff
@@ -1,11 +1,15 @@
...
     args[:icon] = nest(Card.fetch('venn icon'), :view=>:core, :size=>:icon)
   end
 
+  def contribution_count
+    @cc ||= Card.search :type_id=>WikirateAnalysisID,
+              :right_plus=>[ 'article', {:or=>{:created_by=>card.left.name, :edited_by=>card.left.name }}],
+              :return=>:count
+  end
+
+
```
ethn, line 12: nice!

ethn, line 12: :edited_by includes the initial edit (creation).  So we don't really need both created_by and edited_by.

I looked at this query a bit. Looks like it might be possible to speed it up a bit, but I'm not sure what the systemic effects would be.

PR #199, opened 2015-06-14
remove core edit from companies

ethn:
trying to fix a little ugliness on companies and claims by removing code!

### DIFF
--- a/mod/wikirate/set/type/wikirate_company.rb
+++ b/mod/wikirate/set/type/wikirate_company.rb
@@ -13,10 +13,3 @@ def indirect_contributor_search_args
   ]
 end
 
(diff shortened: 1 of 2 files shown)


mod/wikirate/set/type/wikirate_company.rb
```diff
@@ -13,10 +13,3 @@ def indirect_contributor_search_args
   ]
 end
 
-format :html do
-  def edit_slot args
-    # refer to claim.rb
-    super args.merge( :core_edit=>true )
-  end
-end
```
ethn, line 21: this makes you edit within the structure of the page. we're much better off now without this!

PR #200, opened 2015-06-15
Fix wrong url in iframe in source preview page

chuenlok:


Gemfile
```diff
@@ -46,7 +46,7 @@ end
 group :development do
   gem 'rails-dev-tweaks'
   gem 'sprockets' # just so above works
-
+  gem 'thin'
```
chuenlok, line 49: Vignesh found that adding this gem could help the slow problem when running locally. I feel it really helps too. It replaced webrick using another web service called thin.

PR #202, opened 2015-06-18
profile page updates

xithan:


mod/wikirate/set/right/activity.rb
```diff
@@ -36,4 +36,13 @@ def activity_item item
       </div>
     }.html_safe
   end
+
+  view :open do |args|
+    if !card.accountable? && !card.account
```
ethn: I think we only want `!card.account`.

Pretty soon all company cards will be accountable, but we still won't want to show activity unless they actually have an account setup.

ethn: I might put the check in a :when block, btw.

mod/wikirate/set/right/contributed_claims.rb
```diff
@@ -1,15 +1,15 @@
+def contribution_count
+  @cc ||= Card.search :type_id=>ClaimID, :or=>
+                          {:created_by=>cardname.left,
+                           :edited_by=>cardname.left,
```
ethn, line 4: created_by isn't needed (implied by edited_by)

mod/wikirate/set/right/contributed_metrics.rb
```diff
@@ -1,19 +1,20 @@
+def contribution_count
+  @cc ||= Card.search :type_id=>MetricID, :or=>
+                      {
+                       :left=>cardname.left,
```
ethn, line 4: since this `:left=>cardname.left` is in the or clause, won't this find every metric that has a plus card?

mod/wikirate/set/right/follow.rb
```diff
@@ -9,9 +9,14 @@
     card_link card, :text=> 'more...'
   end
 
-  view :profile do |args|
-    frame args.merge(:optional_more_link=>:show) do
-      _render_following_list(args)
+  view :profile, :tags=>:unknown_ok do |args|
+    if !card.accountable? && !card.account
```
ethn: same as above on this one.

mod/wikirate/set/right/my_contributions.rb
```diff
@@ -1,4 +1,37 @@
...
+      {:name=>'Articles', :contributions => :contributed_analysis},
+      {:name=>'Campaigns',:contributions => :contributed_campaigns}
+    ].map do |args|
+      c_card     = user_card.fetch(:trait=>args[:contributions])
+      count      = c_card && c_card.contribution_count
+      contr_name = c_card && c_card.cardname.url_key
+      [ (count || 0), args[:name], contr_name ]
+    end
+  end
+end
```
ethn, line 20: seems like the code got both cleaner and more powerful here!

db/migrate_cards/20150618122941_codename_for_homepage.rb
```diff
@@ -0,0 +1,7 @@
+# -*- encoding : utf-8 -*-
+
+class CodenameForHomepage < Card::Migration
+  def up
+    Card['home'].update_attributes! :codename=>:home
+  end
+end
```
ethn, line 7: "home" is already in use as a codename for the card used to set the homepage:

http://wikirate.org/:home

...so I think this would break.

mod/wikirate/set/right/activity.rb
xithan: views have a :when block?

xithan: > Pretty soon all company cards will be accountable, but we still won't want to show activity unless they actually have an account setup.

That won't happen with the current expression either. But I guess it doesn't happen that a card has a account although it's not accountable.

mod/wikirate/set/right/contributed_metrics.rb
xithan, line 4: It is supposed to find every metric designed by "left". I don't see why it won't do that.

db/migrate_cards/20150618122941_codename_for_homepage.rb
xithan, line 7: ok, I change it to "homepage"

mod/wikirate/set/right/my_contributions.rb
xithan, line 20: ... and less repetitive (had the same searches here as in the "contributed_..." cards)

mod/wikirate/set/right/contributed_metrics.rb
ethn, line 4: OH!!  yeah, this is totally correct. my bad.

PR #204, opened 2015-06-19
Fix url to file

chuenlok:


mod/wikirate_source/set/type/source/source.rb
```diff
@@ -34,7 +34,9 @@ def indirect_contributor_search_args
   end
 end
 
-event :process_source_url, :before=>:process_subcards, :on=>:create, :when=>proc{ |c| not (c.subcards["+File"] or c.subcards["+Text"]) } do
+event :process_source_url, :before=>:process_subcards, :on=>:create, :when=>proc{  |c| 
+  not ( (file_card = c.subcards["+File"] and file_card.stringify_keys.has_key?"content")  or ( text_card =  c.subcards["+Text"] and text_card_content = (text_card.stringify_keys)["content"] and !text_card_content.empty? ))
```
chuenlok: While submitting in web page, the +text and +file which have empty content are also submitted.

xithan: https://github.com/bbatsov/ruby-style-guide#no-and-or-or and https://github.com/bbatsov/ruby-style-guide#safe-assignment-in-condition

xithan: Maybe worth it to put that expression in an extra method to make it more readable. Something like `has_file_or_text?` and then replace the above with `proc { |c| !c.has_file_or_text? }`

PR #210, opened 2015-07-13
fix stack level too deep while rendering article

chuenlok:


mod/wikirate/set/right/wikirate_article.rb
```diff
@@ -24,7 +24,13 @@
   
   def handle_edit_article args
     if params[:edit_article] && card.ok?(:update)
-      render :core, args #to trigger fetching citations information
+      # if missing view renders core view, it will cause infinit loop
+      # core -> missing -> core ... -> stack level too deep
+      if @current_view == :missing 
+        yield
```
ethn: why yield? 

can't we just do

```
render :core, args unless @current_view == :missing
```

?

ethn: looks to me like "yield" will just render the missing view for no reason and then not return the result.

PR #220, opened 2015-07-24
Metric value source

chuenlok:


mod/rating/spec/set/type/metric_value_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    it "updates metric value' value correctly" do
+      @metric = get_a_sample_metric
+      @company = get_a_sample_company
+      subcard = {
+        "+metric"=>{"content"=>@metric.name},
+        "+company"=>{"content"=>"[[#{@company.name}]]",:type_id=>Card::PointerID},
+        "+value"=>{"content"=>"I'm fine, I'm just not happy.", :type_id=>Card::PhraseID},
+        "+year"=>{"content"=>"2015", :type_id=>Card::PointerID},
+        "+Link"=>{:content=>"http://www.google.com/?q=everybodylies", "type_id"=>Card::PhraseID}
+      }
```
ethn, line 13: fyi, we're talking about updating the subcards API.  We should make sure you're up on that conversation.

PR #225, opened 2015-08-05
correct company and contribution counts

xithan:


db/migrate_cards/data/fix_company_counts.json
```diff
@@ -0,0 +1,27 @@
+{
+  "url": null,
+  "timestamp": "2015-08-05 15:25:48 +0200",
+  "card": {
+    "name": "export",
+    "type": "Pointer",
+    "content": "[[metric+company+*type plus right+*structure]]\n[[metric item]]\n[[metric+source+*type plus right++structure]]\n[[metric drag item]]",
+    "value": [
```
ethn, line 8: I think `metric+source+*type plus right++structure` is supposed to be `metric+source+*type plus right+*structure`  (+*structure, not ++structure)

db/migrate_cards/data/fix_company_counts.json
```diff
@@ -0,0 +1,27 @@
...
+      },
+      "",
```
ethn, line 19: ...and that the above typo is why we're getting this empty string here.

PR #239, opened 2015-08-24
fix activity sql error

chuenlok:


mod/wikirate/set/right/activity.rb
```diff
@@ -2,7 +2,7 @@ def virtual?; true end
 
 format :html do
   view :core do |args|
-    Card::Act.where(:actor_id=>card.left.id).order('acted_at DESC').limit(10).map do |act|
+    Card::Act.where("actor_id=#{card.left.id} and card_id is not NULL" ).order('acted_at DESC').limit(10).map do |act|
```
ethn, line 5: we have acts without cards?

ethn:
merged this, but PR is to master, not staging, so it's not showing up as closed.

(still wondering whether master should map to staging server)

ethn:
closing (merged into staging)

PR #244, opened 2015-08-28
Renaming

xithan:
depends on wagn/wagn#491


ethn:
I merged this into staging, but the PR is to master, so it's not showing up as merged.

(might be an argument for making it so that master = dev and somethingelse = live?)

ethn:
closing (merged into staging)

PR #245, opened 2015-08-28
*ltype rtype update for the "_self search change"

xithan:


xithan:
handled by the updated PR wagn/wagn#412

PR #250, opened 2015-09-02
Fix profile page

chuenlok:
1. update some of the test db to fit the test cases. A complete test db will be exported while I finish the new pointer export
2. updated type to type_id in some of the cards
3. correct the search card used after renaming


chuenlok:
it is weird that semaphore is not working suddenly....

PR #252, opened 2015-09-07
test db generator

chuenlok:
A new config file is needed
`config/environment/init_test.rb`
We have `ltype rtype` in wikirate which is not in wagn. Before requiring wikirate mod code, we need to start from raw cardio mod to get and add those needed card from production to test. 

```
Wikirate::Application.configure do
  config.paths["mod"] = "#{Cardio.gem_root}/mod"
end
```


script/metric/import_metric_value.rb
```diff
@@ -11,6 +11,9 @@
     year = row[:year]
     value = row[:value]
     source_url = row[:source]
+    if metric_name.start_with?"EPA"
+      metric_name.gsub!("EPA","United States Environmental Protection Agency")
```
chuenlok: Richard gave me a wrong metric designer name so I try to correct here before importing metric values.

chuenlok:
Really late now. I will check it if some tests fail again tmr

db/migrate_cards/20150907173833_import_production_export.rb
```diff
@@ -0,0 +1,8 @@
+# -*- encoding : utf-8 -*-
+
+class ImportProductionExport < Card::Migration
+  def up
+    import_json "production_export.json"
+    
+  end
+end
```
ethn, line 8: why is this in a migration?  this means all this stuff will get re-imported into the production site when we run migrations (potentially overwriting changes that have happened in the meantime)

mod/wikirate/set/type/pointer/pointer.rb
```diff
@@ -0,0 +1,76 @@
+view :solocomplete do |args|
```
ethn: final filename should be more descriptive

mod/wikirate/set/type/wikirate_company.rb
```diff
@@ -7,10 +7,10 @@
...
   [
-    {:type_id=>ClaimID, :plus=>['company',:link_to=>self.name]},
-    {:type=>'source', :plus=>['company',:link_to=>self.name]},
-    {:type=>'analysis', :left=>self.name },
-    {:type=>'metric value', :left=>{:right=>self.name}}
+    {:type_id=>Card::ClaimID, :plus=>['company',:link_to=>self.name]},
+    {:type_id=>Card::SourceID, :plus=>['company',:link_to=>self.name]},
+    {:type_id=>Card::WikirateAnalysisID, :left=>self.name },
+    {:type_id=>Card::MetricValueID, :left=>{:right=>self.name}}
   ]
```
ethn, line 14: fwiw, Card:: should be unnecessary here (no big deal, it can stay)

ethn, line 14: happy you're making these changes!

script/metric/import_metric_value.rb
ethn: so the idea is that this is a short-term fix?

I gather that it's not easy to fix in the data?

mod/wikirate/set/type/pointer/pointer.rb
ethn: I assume you were having some sort of problem with the old name?

ethn:
seems like a massive improvement; glad to see all those codenames!!

lib/tasks/wikirate.rake
```diff
@@ -16,136 +16,68 @@ namespace :wikirate do
       mysql_args += " -p #{pwd}" if pwd
       system "mysql #{mysql_args} #{test_database} < #{db_path}"
     end
+    desc 'add wikirate test data to test database'
+    task :add_wikirate_test_data do
```
ethn, line 21: vast improvement.  still seems like a pretty long task; would be nice to separate the dumping into its own task (eg if there were a testing problem you could try migrating an existing database and then dumping)

ethn:
....also nice to get rid of that huge sql dump!

db/migrate_cards/20150907173833_import_production_export.rb
chuenlok, line 8: this is just to create three cards.
1. `production_export` pointer card that contains all needed card names only
2.  two search cards that get all cards with codename and type as cardtype

This will not affect the current cards as I know.

mod/wikirate/set/type/wikirate_company.rb
chuenlok, line 14: I don't know when I should/shouldn't include `Card::`. It works occasionally so I add `Card::` everywhere.

xithan, line 14: Just found two places :)  I needed `Card::` in a rake task and in `Card::Log::Performance`

ethn, line 14: You don't need it in set modules, because when the module is created, the code is in:

```
class Card
  module Set
    module Type
    ....
    end
  end
end
```

In other words, you're already in the context of `Card`.  If you're outside of that context (eg in a rake task), you need the full name.

mod/01_rating/spec/set/right/downvotee_search_spec.rb
```diff
@@ -24,7 +24,7 @@
           vcc2.save!
         end
         metric_downvotee_search_card = Card.fetch "#{apple.name}+metric+downvotee search"
-        result = metric_downvotee_search_card.item_cards
+        result = metric_downvotee_search_card.format.get_search_result
```
ethn, line 27: what is this change about?

lib/tasks/wikirate.rake
```diff
@@ -17,134 +17,82 @@ namespace :wikirate do
...
+              puts "creating card #{c} #{Card.create!(c)}".yellow
+            end
+          rescue => e
+            puts "Error in #{c} #{e}".red
           end
         end
+        FileUtils.rm_rf(Dir.glob('tmp/*'))
+        seed_test_db = "RAILS_ENV=test rake wikirate:test:add_wikirate_test_data"
+        puts seed_test_db.green
+        system seed_test_db
```
ethn, line 76: I suspect we will want to run migrations here.

hmmm..  but we won't know what migrations have been run, huh?

ethn, line 76: eg, the latest wagn master includes significant migrations.  

I think I'm going to go ahead and merge this without the latest wagn so we can get the special pointer ready on dev at least.  But we need to think about how to handle these migration situations.

mod/01_rating/spec/set/right/downvotee_search_spec.rb
chuenlok, line 27: `item_cards` is do the raw query which won't follow the order of votes. 

```
def item_names params={}
  Card.search(query(params.merge(:return=>:name)))
end
```

`get_search_result` will re-order according the users' voting history. 

view `card_list` will render the result using `search_results` which will include the re-ordering.

lib/tasks/wikirate.rake
chuenlok, line 76: This part of the migration is adding some hard-coded test data only. If the test db starts from the raw wagn seed db, we need it to add some sample companies, sources. I don't think it is related to the migration version problem.

mod/01_rating/spec/set/right/downvotee_search_spec.rb
ethn, line 27: that makes sense, but I think we might be able to organize this a bit more clearly:
1. get_search_result should have a clearer name, and
2. it should probably be a method on the model, not the format.  It's much more natural for the format object to call this (it already has a card object) rather than have the model create a format object just for this purpose (card.format creates a new object)

PR #257, opened 2015-09-16
remove non exist tables in export

chuenlok:
1. I just found that `schema_migrations_cards` does not exist anymore but I included it to the migration export. It is removed now.
2. Philipp fixed `_self/_right` stuff, so the queries are needed to be updated. (the migration won't take care of the things in code)


ethn:
excellent.  Thank you, henry.

PR #262, opened 2015-09-29
handle some exception

chuenlok:
related to https://wagn.airbrake.io/projects/95878/groups/1518205809789004593 , http://dev.wikirate.org/Clean_Clothes_Campaign+contribution 
As `Clean_Clothes_Campaign` does not exist, all things throw exception while using the left part of the card, eg, `card.left.account`.

### DIFF
--- a/mod/wikirate/set/right/my_contributions.rb
+++ b/mod/wikirate/set/right/my_contributions.rb
@@ -4,17 +4,19 @@ end
 
 def sections
   @sections ||= begin
-    user_card = left
-    [ {:name=>Card[:metric].name,  :contributions => :contributed_metrics},
-      {:name=>Card[:claim].name,   :contributions => :contributed_claims},
-      {:name=>Card[:source].name,  :contributions => :contributed_sources},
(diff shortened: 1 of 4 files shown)


mod/wikirate/set/right/my_contributions.rb
```diff
@@ -22,15 +24,19 @@ def sections
...
-      section_args = {:view=>:open, :title=>name, :hide=>'menu'}
-      # FIXME - cardname
-      if name == 'Initiative'
-        nest Card.fetch(contr_name), section_args.merge(:item=>{:view=>:content, :structure=>'initiative item'})
-      else
-        nest Card.fetch(contr_name), section_args
-      end
-    end.join "\n"
+    if card.sections
+      card.sections.sort.reverse.map do |count, name, contr_name|
```
chuenlok, line 30: puppy, you have to check correctly.
count is used in the block.

ethn, line 30: where?

PR #265, opened 2015-10-16
Merge subcards api

chuenlok:
need latest subcards fix from Philipp


mod/01_rating/set/type/metric_value.rb
```diff
@@ -46,14 +49,27 @@ def source_exist?
 
 def create_source
   Env.params[:sourcebox] = 'true'
-  value = subcards.delete('+value')
-  source_card = Card.create :type_id=>Card::SourceID, :subcards=>subcards.clone
+  value = subfield 'value'
+  remove_subfield 'value'
```
xithan: I improved the API. You can use `value = remove_subfield 'value'` now.

mod/01_rating/set/type/metric_value.rb
```diff
@@ -46,14 +49,27 @@ def source_exist?
...
+  _subcards = {}
+  ['file', 'link', 'text'].compact.each do |key|
```
xithan: no nils here for `compact`

mod/01_rating/set/type/metric_value.rb
```diff
@@ -46,14 +49,27 @@ def source_exist?
...
+      if key =='file'
+        _subcards["+#{key}"] = { file: subcard.file.file, type_id: subcard.type_id }
+      else
+        _subcards["+#{key}"] = { content: subcard.content, type_id: subcard.type_id }          
+      end
+    end
+  end
+  subcards.each_with_key do |subcard, key|
+    remove_subcard key
+  end
```
xithan: there is `clear_subcards` now for this

chuenlok:
I haven't touched the config file for ruby style but it fails suddenly. I tried the file with rubocop locally. It runs normally.

mod/01_rating/spec/set/right/downvotee_search_spec.rb
```diff
@@ -121,7 +121,16 @@
...
+              "+source"=>{
+                "subcards"=>{
+                  "new source"=>{
+                    "+Link"=>{
+                      "content"=>"http://www.google.com/?q=fringepeter",
+                       "type_id"=>Card::PhraseID
+                    }
+                  }
+                }
+              }
```
ethn, line 133: why does the old API not work here any more?

mod/wikirate/set/type_plus_right/metric_value/source.rb
```diff
@@ -2,7 +2,7 @@
   view :editor do |args|
     
     if card.new?
-      source = Card.new :type_code=>:source
+      source = Card.new type_code: :source, name: "new source"
```
ethn, line 5: So could I break the site if I created a new Basic card called "new source"?

mod/01_rating/spec/set/right/downvotee_search_spec.rb
ethn, line 133: the new API was supposed to clean things up, but this PR adds a lot more than it removes.  And this now appears to skip the "+Link" handling that is still in the code, no?

chuenlok, line 133: > the new API was supposed to clean things up, but this PR adds a lot more than it removes.

The new API helps standardize the use of subcards as an object which help the readability of using subcards. As subcards is not a hash anymore, the use of subcards in wikirate may need more lines of code to fit into new subcards API. Before new API, subcards' hash is straight forward and no context things are considered. It is easy to modify and extract things in the hash. For example, I can use `clone` to get the params and create a new card which requires more effort in the new API. I can say the new API clean things up by standardizing things but not greatly reducing lines of code. 
Moreover, as the houndci requires a lot of style changes so things are broken into lines to pass the houndci checking. 

>  And this now appears to skip the "+Link" handling that is still in the code, no?

This part is trying to simulate the request what will be sent from the browser.

If you remember, the metric value creation page renders the new source editor to let users input source information. Before the new subcard API, the rendering was just straight forward. For example, the link input filed's name was just `card[subcards][+Link][content]` but now the name now is `card[subcards][+source][subcards][new source][subcards][+Link][content]`.
from `form.rb`

``` ruby
def subcard_input_names
    return '' if !form_root_format || form_root_format == self
    "#{@parent.subcard_input_names}[subcards][#{card.contextual_name}]"
  end
```

I have to put a name to the new source card while rendering otherwise the name will be `card[subcards][+source][subcards][][subcards][+Link][content]`  which is treated as an array but not a subcard.

`+Link` handling is not in the code anymore now. Please refer to `type/metric_value.rb#create_source`. Instead, a source card will be created based on the subcards in the `+source`.

mod/wikirate/set/type_plus_right/metric_value/source.rb
chuenlok, line 5: Newing a card with exiting card name should not cause any problem because it will not cause conflicts until we really save it.

mod/01_rating/spec/set/right/downvotee_search_spec.rb
ethn, line 133: Very thorough answer, Henry. Thank you.

I guess I was looking at old code on the +Link issue.

It seems to me like the need to add a fake name ("new source") for the API to get things to work is a wart.  We probably need some standard way to handle autonamed cards. (eg _autoname_1).  But the "new source" workaround makes sense for now. (In general, it's best to use a "_" in front of temporary names just to reduce the chance of conflict).

If I understand correctly, the added complexity here is really about the fact that our API can now handle trees.  Before, if we had a form within a form within a form, we would effectively compress them all so that all the cards were subcards of one parent.  Now there can be many generations.  This is really powerful, because it means, for example, that you could create a setup where the +Link is required for a Source but a source is not required for a Note (or similar custom rules).

Thanks for clarifying.  I'm going to merge and deploy this first so that it's easier to review and test the other PR.

mod/wikirate/features/metric_value.feature
```diff
@@ -13,7 +13,7 @@ Feature: note
     And I fill in "pointer_item" with "Death Star" within "form > fieldset.editor > .RIGHT-company"
     And I fill in "pointer_item" with "2015" within "form > fieldset.editor > .RIGHT-year"
     And I fill in "card_subcards__value_content" with "101"
-    And I fill in "card_subcards__Link_content" with "http://example.com"
+    And I fill in "card_subcards__source_subcards_new_source_subcards__Link_content" with "http://example.com"
     And I press "Submit"
```
ethn, line 17: in concept, these cukes are supposed to use user-facing terms (which might mean we should add labels?)

PR #269, opened 2015-11-05
fix error in analysis page

chuenlok:


mod/wikirate/set/all/wikirate.rb
```diff
@@ -212,7 +212,7 @@ def main_name
   end
 
   def main_type_id
-    @main_type_id ||= Card[main_name].type_id
+    @main_type_id ||= Card.fetch(main_name).type_id
```
ethn, line 215: should probably use :skip_modules to avoid the unnecessary loading.  not a big deal.

PR #270, opened 2015-11-06
trigger houndci

chuenlok:


ethn:
so we can delete this now, right?

chuenlok:
yes

PR #276, opened 2015-11-09
Metric button from source

chuenlok:


mod/wikirate/set/type_plus_right/metric_value/source.rb
```diff
@@ -1,7 +1,6 @@
 format :html do
   view :editor do |args|
-    
-    if card.new?
+    if card.new? && !(Env.params['slot'] && Env.params['slot']['source'])
```
ethn: couldn't this test just be args['source'] ?

mod/01_rating/spec/set/type/metric_value_spec.rb
```diff
@@ -54,6 +54,31 @@
       value_card = Card["#{@metric_value.name}+value"]
       expect(value_card.content).to eq("I'm fine, I'm just not happy.")
     end
+
+    it 'creates metric value with source' do
```
ethn: does this mean it's currently possible to create a metric value without a source?

PR #278, opened 2015-11-12
fix source adding and import redirection

chuenlok:
updated test db from production


mod/wikirate_source/set/type_plus_right/source/file/import.rb
```diff
@@ -68,7 +68,11 @@
       end
     end
     if errors.empty?
-      abort :success=>"REDIRECT: #{metric_pointer_card.item_names.first}" 
+      abort success: {
+        name: metric_pointer_card.item_names.first,
+        redirect: true,
+        view: :open
+      }
```
ethn, line 75: this event REALLY needs to be broken down into smaller methods.

(the specific code change is an improvement, though!)

mod/wikirate_source/features/source.feature
```diff
@@ -0,0 +1,36 @@
...
+  I want to create a source
+
+  Background:
+    Given I am signed in as Joe Admin 
+
+  Scenario: Create a link source
+    When I wait until ajax response done
+    And I go to new Source
+    And I fill in "card_subcards__Link_content" with "http://example.com"
+    And I fill in "card_subcards__Title_content" with "a test link source"
```
ethn, line 14: nice test.  We need to work on these "fill in" steps; cukes aren't supposed to have code identifiers.  Maybe we need to add labels or something?

PR #288, opened 2015-12-01
fix adding metric value

chuenlok:


wagn:
there's lots of stylistic cleanup here; can you point me to the actual fix?

chuenlok:
on line 70 of `mod/01_rating/set/type/metric_value.rb`. 

I need to approve subcards so that I can have the file in the cache.

I think I will do a style clean up on the whole project after the review.

ethn:
ok thanks (inline comments in the code are preferred)

PR #289, opened 2015-12-02
migrate tags to topic, update analysis_contrubtion and cached expiration

chuenlok:


mod/01_rating/set/right/analysis_contributions.rb
```diff
@@ -11,31 +33,41 @@ def data_item content, label, type=:default
...
-    source_cnt = (sources = Card.fetch("#{analysis_name}+sources")) && sources.cached_count
+  view :core do |_args|
+    analysis_article_name = "#{analysis_name}+#{Card[:wikirate_article].name}"
+    overview_card = Card.fetch analysis_article_name
+    analysis_claim_name = "#{analysis_name}+#{Card[:claim].name}"
+    claim_cnt = (claims = Card.fetch(analysis_claim_name)) &&
+                claims.cached_count
+    source_cnt = (sources = Card.fetch("#{analysis_name}+sources")) &&
+                 sources.cached_count
+    metric_cnt = (metrics = Card.fetch("#{analysis_name}+metrics")) &&
```
chuenlok, line 49: added count for `analysis+metric` search.

mod/wikirate/set/type/claim.rb
```diff
@@ -77,7 +77,7 @@ def edit_slot args
     # special view for prompting users with next steps
     if Auth.signed_in? and ( tip = args[:tip] || next_step_tip ) and @mode != :closed
       %{
-        <div class="claim-tip">
+        <div class="note-tip">
```
chuenlok, line 80: Change the wordings from `claim` to `note`

mod/caching/set/type_plus_right/wikirate_analysis/claim.rb
```diff
@@ -1,8 +1,17 @@
...
+     changed_card.type_code == :pointer
+    # find all related analysis
+    card_names = changed_card.item_names.unshift('in')
+    analysis_type_id = Card::WikirateAnalysisID
+    case r.key
+    when 'company'
+      Card.search type_id: analysis_type_id, left: card_names, append: 'claim'
+    when 'topic'
+      Card.search type_id: analysis_type_id, right: card_names, append: 'claim'
+    end
```
xithan: You can reduce repetition if you set only :left/:right in the case statement and call Card.search after that. Makes it easier to spot the difference.

mod/caching/set/type_plus_right/wikirate_analysis/metric.rb
```diff
@@ -0,0 +1,33 @@
...
+                      append: 'metric' }
+      if (topic_card = ll.fetch trait: :wikirate_topic) &&
+         (topic_cards = topic_card.item_names) && topic_cards.size > 0
+        search_args.merge(right: topic_cards.unshift('in'))
+      end
+      expired_cached_cards.concat(Card.search search_args)
+    end
+  end
+  expired_cached_cards.uniq
+end
```
xithan, line 37: because we execute this for every card I suggest you add as first line for better performance
`return if changed_card.simple? || (changed_card.right_id != CachedCountID && changed_card.type_id != PointerID)`
This will reject most of the cards without a card lookup

mod/wikirate/set/type/claim.rb
xithan, line 80: did you check that there are no css rules that refer to these claim classes?

mod/caching/set/type_plus_right/wikirate_analysis/source.rb
```diff
@@ -1,8 +1,17 @@
...
+     changed_card.type_code == :pointer
+    # find all related analysis
+    card_names = changed_card.item_names.unshift('in')
+    analysis_type_id = Card::WikirateAnalysisID
+    case r.key
+    when 'company'
+      Card.search type_id: analysis_type_id, left: card_names, append: 'source'
+    when 'topic'
+      Card.search type_id: analysis_type_id, right: card_names, append: 'source'
+    end
```
xithan: same repetition as above. The hound would complain if he were smart enough ;)

mod/wikirate/set/type/claim.rb
chuenlok, line 80: Vignesh requests changing this to fit correct css class.

mod/caching/set/type_plus_right/wikirate_analysis/claim.rb
```diff
@@ -1,8 +1,19 @@
...
+    card_names = changed_card.item_names.unshift('in')
+    analysis_type_id = Card::WikirateAnalysisID
+    args = { type_id: analysis_type_id, append: 'claim' }
+    case r.key
+    when 'company'
+      args.merge!(left: card_names)
+    when 'topic'
+      args.merge!(right: card_names)
+    end
+    Card.search args
```
xithan: what about one step further? 

```
side = case r.key
       when 'company' then :left
       when 'topic' then :right
       end
Card.search type_id: Card::WikirateAnalysisID, append: 'claim',  side => card_names
```

mod/caching/set/type_plus_right/wikirate_analysis/metric.rb
```diff
@@ -0,0 +1,37 @@
+include Card::CachedCount
+
+expired_cached_count_cards do |changed_card|
+  # "right_plus":[["topic",{"refer_to":"_2"}],
+  #  ["_ll",{"right_plus":["*cached count",{"ne":0}]}] ]
+
+  unless changed_card.simple? || (changed_card.right_id != CachedCountID &&
+                                  changed_card.type_id != PointerID)
```
xithan, line 8: You can use `next` in a block as guard clause. That would eliminate the codeclimate warning.

PR #293, opened 2015-12-04
migration for tags to topics

chuenlok:


ethn:
merging this without really studying it.  might be worth discussing optimization with Philipp tomorrow.

PR #294, opened 2015-12-04
Fix updating metric values with another source, export from dev

chuenlok:


db/migrate_cards/20151202112547_import_migrate_tags_to_topics.rb
```diff
@@ -6,16 +6,16 @@ def put_things_in_tag_to_correct_position cards, skip_year
     company = card.fetch trait: :wikirate_company, new: {}
```
ethn, line 7: I suggest clearing the cache (Card.cache.reset_local, I think) after, say, every 10 cards.  I suspect we're loading too many cards into the local cache, and that's causing things to bog down.

Can we do that and try to deploy to see how it goes?

PR #301, opened 2015-12-16
contribution and cached_count cleaning

ethn:
the biggest thing here is that I deleted something I hope isn't important (see comments)


mod/caching/set/type_plus_right/wikirate_analysis/metric.rb
```diff
@@ -11,14 +11,15 @@
...
-end
+
+# expired_cached_count_cards(
+#   set: TypePlusRight::Metric::WikirateTopic
+# ) do |changed_card|
+#   card_names = changed_card.item_names.unshift('in')
+#   Card.search type_id: Card::WikirateAnalysisID,
+#               right: card_names,
+#               append: 'metric'
+# end
```
ethn: So, I'm hoping the above is unnecessary, because it takes metric creation from a couple of seconds to a couple of minutes.

ethn: Do we need (analysis)+metric+*cached_count?

mod/caching/set/all/contributions.rb
```diff
@@ -1,63 +1,75 @@
...
+  indirect_contributor_search_args.inject([]) do |cards, search_args|
+    cards + Card.search(search_args)
+  end
+end
+
+def calculate_indirect_contribution_count
+  indirect_contributor.inject(0) do |total, c_card|
+    more_contributions =
+      case c_card
+      when respond_to?(:contribution_count)
```
xithan: This doesn't work. It's equivalent to `c_card === respond_to?(:contribution_count)` which is always false.

mod/01_rating/set/type/metric.rb
```diff
@@ -17,3 +17,14 @@ def view_caching?
     true
   end
 end
+
+def analysis_names
```
ethn, line 21: same code as in source and claim; should be shared.

mod/caching/set/type_plus_right/metric/all_values.rb
```diff
@@ -2,38 +2,18 @@
...
 
-# expire if company name was changed
-ensure_set { Type::MetricCompany }
-expired_cached_count_cards set: Type::MetricCompany,
-                           changed: :name do |changed_card|
-  metrics = Card.search type_id: MetricID, right_plus: changed_card.name
-  metrics.map do |metric|
-    metric.fetch trait: :all_values
-  end
-end
```
ethn, line 30: I deleted this one because: 
(a) if a company name changes, then that automatically renames all its descendants, including analyses and values, so this isn't really necessary, and
(b) MetricCompany isn't anything anyway, right?

mod/caching/set/type_plus_right/metric/all_values.rb
```diff
@@ -2,38 +2,18 @@
...
-
-# initialize cached values hash if new metric is created
-expired_cached_count_cards set: Type::Metric, on: :create do |changed_card|
-  changed_card.fetch trait: :all_values
-end
```
ethn, line 35: I think deleting `def cached_count` above means this will be generated lazily, no?

mod/caching/set/type_plus_right/metric/wikirate_company.rb
```diff
@@ -1,24 +1,10 @@
...
-    changed_card.metric_card.fetch(trait: :wikirate_company)
-  ]
-  # it is to refresh the related metric value to the company
-  # refer to ltype_rtype/metric/wikirate_company.rb
-  if (metric_card = changed_card.metric_card) &&
-     metric_card.type_id == MetricID &&
-     (company_card = changed_card.company_card) &&
-     company_card.type_id == WikirateCompanyID
-    result.push(Card.fetch("#{metric_card.name}+#{company_card.name}"))
-  end
```
ethn, line 22: Frankly, I couldn't make sense of this so I deleted it.  the result isn't even a search card, so this appears to be a hack

ethn, line 22: also, this is not a good place to validate the types of the parts of a metric value

mod/caching/set/type_plus_right/wikirate_company/analyses_with_articles.rb
```diff
@@ -1,12 +1,12 @@
...
-expired_cached_count_cards(
-  set: TypePlusRight::WikirateAnalysis::WikirateArticle,
-  on: [:create, :delete]
-) do |changed_card|
-  (analysis = changed_card.left) && (company = analysis.left) &&
-    company.fetch(trait: :analyses_with_articles)
+recount_trigger(TypePlusRight::WikirateAnalysis::WikirateArticle,
+                on: [:create, :delete]) do |changed_card|
+  if (company_name = changed_card.cardname.left_name.left)
+    Card.fetch company_name.to_name.trait(:analyses_with_articles)
```
ethn, line 10: there are lots of places where I was able to get rid of a lot of card fetching by using name methods.

mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
```diff
@@ -1,23 +1,17 @@
...
-expired_cached_count_cards set: TypePlusRight::Claim::WikirateCompany do |changed_card|
+# recount topics associated with a company whenever <note>+company is edited
+recount_trigger Claim::WikirateCompany do |changed_card|
   changed_card.item_names.map do |company_name|
-    Card.fetch "#{company_name}+#{Card[:wikirate_topic].name}"
+    Card.fetch company_name.to_name.trait(:wikirate_topic)
   end
 end
+
+# FIXME: should also count connections via metrics.
```
ethn: notable fixme.

mod/caching/set/type_plus_right/wikirate_analysis/claim.rb
```diff
@@ -1,17 +1,21 @@
 include Card::CachedCount
 
+def self.notes_for_analyses_applicable_to note
+  note.analysis_names.map do |analysis_name|
+    Card.fetch analysis_name.to_name.trait(:claim)
+  end
+end
```
ethn, line 7: defining this exact method three times. should find one place for it.

mod/caching/set/type_plus_right/wikirate_analysis/claim.rb
```diff
@@ -1,17 +1,21 @@
...
 
+# ... <note>+topic is edited
 ensure_set { TypePlusRight::Claim::WikirateTopic }
-expired_cached_count_cards set: TypePlusRight::Claim::WikirateTopic do |changed_card|
-    # find all related analysis
-    card_names = changed_card.item_names.unshift('in')
-    Card.search type_id: Card::WikirateAnalysisID, append: 'claim',
-                right: card_names
+recount_trigger TypePlusRight::Claim::WikirateTopic do |changed_card|
+  notes_for_analyses_applicable_to changed_card.left
```
ethn, line 20: so my concern about this mod (and the two others just like it) is that it seems like it will recount analyses twice if someone edits +company and +topic in the same submission (might be common on create, for example).

mod/caching/set/type_plus_right/metric/wikirate_company.rb
xithan, line 22: This was on my list to be refactored. I don't think we need the last part. The result has a calculate_count method so I wouldn't call it a hack but the cache update is handled in `ltype_rtype/metric/wikirate_company.rb`

PR #310, opened 2016-01-05
update the vote count type to number

chuenlok:
The wagn way to update take too much time related to the cache. I think updating the type id directly should be fine.


ethn:
This is fine; should probably add a global cache reset to the end.

fwiw, you could do this even faster with a single msyql query (update cards set type_id = X where right_id = Y)

ethn:
looks good. is it safe to assume it's been tested?

chuenlok:
Tested locally. I did not see anything goes wrong at the moment.

PR #311, opened 2016-01-06
Rename about

chuenlok:
1. updated the card name of the abouts card of companies and topics.
2. update the `company/topic+right sidebar+*type plus right+*structure`
3. rename the company+about source file name
4. remove the add general overview link

### DIFF
--- a/db/migrate_cards/20160106014913_import_company_topic_right_sidebar_structure.rb
+++ b/db/migrate_cards/20160106014913_import_company_topic_right_sidebar_structure.rb
@@ -2,7 +2,6 @@
 
 class ImportCompanyTopicRightSidebarStructure < Card::Migration
   def up
-    import_json "company_topic_right_sidebar_structure.json"
-    
+    import_json 'company_topic_right_sidebar_structure.json'
   end
 end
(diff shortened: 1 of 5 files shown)


mod/wikirate/set/type_plus_right/wikirate_company/general_overview.rb
```diff
@@ -0,0 +1,4 @@
+
+view :missing do |_args|
+  link_to 'Add General Overview', { view: :new }, remote: false
+end
```
ethn: why do we need this?

db/migrate_cards/20160106014913_import_company_topic_right_sidebar_structure.rb
```diff
@@ -0,0 +1,8 @@
+# -*- encoding : utf-8 -*-
+
+class ImportCompanyTopicRightSidebarStructure < Card::Migration
+  def up
+    import_json "company_topic_right_sidebar_structure.json"
```
ethn: is your editor notifying you of these problems?

mod/wikirate/set/type_plus_right/wikirate_company/general_overview.rb
chuenlok: I remember at some moment we wanted to have the add about to link to a new card page instead of creating inside the slot. 

Do we still want to keep that?

ethn: no, it needs to happen in place.

ethn: (that must have been for when About was at the top)

PR #315, opened 2016-01-07
citation box fix

xithan:


mod/wikirate/set/right/citation_count.rb
```diff
@@ -32,6 +39,6 @@ def no_citations? args
   end
 
   def related_overview_card
-    Card.fetch "#{card.cardname.left}+#{Card[:related_articles]}"
+    Card.fetch "#{card.cardname.left}+#{Card[:related_articles].name}"
```
ethn: my favorite syntax for this (and actually slightly faster):
`Card.fetch card.cardname.left_name.trait(:related_articles)`

ethn: I suppose we could make it even tighter if Card::Name had a .card method :)

xithan: I forgot that Name has also a trait method and I didn't want to call `card.left.trait` because I don't know if `left` exists.

PR #316, opened 2016-01-07
General Overview 

srivig:


mod/wikirate/set/right/citation_count.rb
```diff
@@ -20,10 +20,17 @@ def default_related_overview_modal_box_args args
         text: link_text,
         html_args: { class: "#{'no-citations' if no_citations?(args)}" }
       )
-    return unless parent && (pp = parent.parent) && (ppl = pp.card.left) &&
-                  ppl.type_id == WikirateCompanyID
+    return unless (company = on_company_page?)
     link_args[:path_opts] ||= {}
-    link_args[:path_opts][:company] = ppl.key
+    link_args[:path_opts][:company] = company.key
+    link_args[:path_opts][:general_overview] = true
```
ethn, line 26: I'm glad you added the `on_company_page?` test, but this still has a ways to go to be clean.

what if we're on a topic page?  that should be general_overview too, no?

mod/wikirate/set/right/citation_count.rb
```diff
@@ -20,10 +20,17 @@ def default_related_overview_modal_box_args args
...
+  end
+
+  def on_company_page?
+    return unless parent && (pp = parent.parent) && (ppc = pp.card)
+    return ppc if ppc.type_id == WikirateCompanyID
+    return unless (ppl = ppc.left) && ppl.type_id == WikirateCompanyID
+    return ppl
```
ethn: we need better variable names.

more importantly, this code looks really fragile to me.

mod/wikirate/set/right/general_overview.rb
```diff
@@ -0,0 +1,36 @@
+format :html do
+  view :content do |args|
+    handle_edit_general_overview( args ) { super args }
```
ethn: we need to address these hound comments, please.

mod/wikirate/set/right/general_overview.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  view :missing do |args|
```
ethn: you could just use `view :missing, :content`

PR #318, opened 2016-01-08
Create gibs user

chuenlok:
Richard wants to create a bunch of users for gibs so I created this script.

As the CSV file contains the users' information, I will manually upload it to live/dev and run the script once it is deployed.

As we may not want the users receiving too many emails, so I redirect all email to a dummy email receiver service.


ethn:
better to turn off deliveries than redirect.  redirecting means we still have to do all the processing.

PR #321, opened 2016-01-09
Tabs fix

chuenlok:
1. updated the create gibs users script. Richard wants to send another welcome emails separately.
2. latest export from dev
3. added URL modifying function. I tested it with an IE8 and it seems working. I am using the IE8 emulator in IE Edge which I am not sure if it is a real IE 8.
4. tabs can be cmd/ctrl + click now
5. only the left sidebar content is loaded in ajax


mod/wikirate/set/right/tab_content.rb
```diff
@@ -0,0 +1,19 @@
...
+    # show the content based on the url parameter
+    # tabs: metric, topic, company, note, reference, overview
+    tab = Env.params['tab']
+    left_name = card.cardname.left
+    card_tab_name =
+      if !tab.nil?
+        "#{left_name}+#{tab}_page"
+      else
+        "#{left_name}+metric_page"
+      end
```
ethn, line 12: could simplify to:

```
tab ||= 'metric'
card_tab_name = "#{left_name}+#{tab}_page"
```

mod/wikirate/set/right/tab_content.rb
```diff
@@ -0,0 +1,19 @@
+format :html do
+  view :core do |_args|
+    # show the content based on the url parameter
+    # tabs: metric, topic, company, note, reference, overview
+    tab = Env.params['tab']
+    left_name = card.cardname.left
```
ethn, line 6: +1 for doing this with names!

mod/wikirate/set/right/tab_content.rb
```diff
@@ -0,0 +1,19 @@
...
+    card_tab_name =
+      if !tab.nil?
+        "#{left_name}+#{tab}_page"
+      else
+        "#{left_name}+metric_page"
+      end
+    if (content_card = Card.fetch card_tab_name)
+      subformat(content_card).render_content
+    else
+      ''
```
ethn, line 16: I might put an error like "no such tab: #{tab}".  This should only happen if we have a bad tab name, right?

script/create_gibs_users.rb
```diff
@@ -14,12 +14,10 @@
...
+    Card.create! args
     puts "activating #{row[:name]}"
     status_card = Card["#{row[:name]}+*account+*status"]
     status_card.update_attributes! content: 'active', silent_change: true
-
-    Card::Mailer.perform_deliveries = true
-    user.account.send_welcome_email
   end
 end
+Card::Mailer.perform_deliveries = true
```
ethn, line 23: fwiw, this is probably unnecessary, because the setting is specific to the process running this script.

mod/wikirate/set/right/tab_content.rb
ethn, line 12: or even:

```
tab = Env.params['tab'] || 'metric'
card_tab_name = "#{card.cardname.left}+#{tab}_page"
```

PR #338, opened 2016-01-18
fix editing metric value

chuenlok:


mod/01_rating/set/type/metric_value.rb
```diff
@@ -20,13 +20,17 @@ def company_card
...
+def source_subcards new_source_card
+  [new_source_card.subfield :file, new_source_card.subfield :text,
+    new_source_card.subfield :wikirate_link]
+end
+
 def source_exist?
   sub_source_card = subfield('source')
-  return false if sub_source_card.nil?
+  return false if sub_source_card.nil? ||
+                  sub_source_card.subcard('new_source').nil?
```
chuenlok, line 31: when editing metric value, `new source` does not exist in subfield source

ethn, line 31: doesn't this mean sources are no longer required?

(I didn't really intend to merge this yet.  doh.)

ethn, line 31: what is 'new source'?  this check is only called when editing a metric value, right?

chuenlok, line 31: `new source` is the source card we used to render the editor in the new metric value form. This would be replaced with new source box later.

I keep this `source` because the source will be updated when someone is trying to update the metric value in metric CSV import. We need to check whether it is a normal update or import update.

PR #339, opened 2016-01-25
Fix general overview, new metric value form

chuenlok:
For new metric value form, I added metric field to the structure and render when it is not from metric or company.


ethn:
not sure I understand it all, but I'll trust you.

it would be good to do some hard qa testing on metric value adding.  Like, try your hardest to break it as a user with weird values, missing values, etc.  eg I think Richard may have broken it by starting a source website with "www" rather than "http".

when you're confident, you can merge and deploy.

chuenlok:
I will add more test cases before merging it.

mod/01_rating/set/type/metric_value.rb
```diff
@@ -106,31 +106,40 @@ def create_source
 end
 
 format :html do
-  def default_new_args args
+  def get_structure args
```
chuenlok, line 133: What structure will be used depends on the context. If a new metric value from is with company in the argument, a metric value form without a company field will be shown and so on.

We did not make a correct structure for a default new metric value form so I created a new structure with all fields. It will be shown when there is no company/source/metric in the argument.

mod/01_rating/set/type/metric_value.rb
```diff
@@ -106,31 +106,40 @@ def create_source
...
+    elsif args[:metric]
+      'metric add value'
+    else
+      'default add metric value'
+    end
+  end
+
+  def set_hidden_args args
     if !args[:source]
+      view = (args[:metric] || args[:company]) ? :titled : :open
```
chuenlok, line 147: the redirection should a normal open view if the metric value is created on a new default metric value form.

mod/wikirate/set/right/general_overview.rb
```diff
@@ -3,28 +3,38 @@
     handle_edit_general_overview(args) { super args }
   end
 
-  view :content do |args|
+  view :core do |args|
```
chuenlok, line 6: these views will show a editor even the card is not in edit mode.

While citation button is clicked, we only need to show editor with tips to copy of the general overview section. Other things in the company page should be shown as normal. 

missing and core view should be overridden for this case.

PR #348, opened 2016-02-17
migration to rename

chuenlok:


db/migrate_cards/data/renaming_initiative.json
```diff
@@ -0,0 +1,31 @@
...
+  "card": {
+    "name": "export",
+    "type": "Pointer",
+    "content": "[[project+open]]\n[[project+Closed]]\n[[project+*self+*structure]]\n[[nav bar menu]]",
+    "value": [
+      {
+        "name": "Project+Open",
+        "type": "Search",
+        "content": "{ \r\n  \"type\":\"Project\",\r\n  \"right_plus\":[\"status\",{\"refer_to\":\"_right\"}]\r\n}"
+      },
```
ethn, line 13: Doesn't renaming "Initiative" to "Project" automatically do the above?

chuenlok, line 13: It will rename the card name but not the content. Some of them have `[[Initiative]]` which are not updated automatically.

ethn, line 13: Interesting.  I wonder if that's connected to my new reference handling.  I'd kind of like to fix this now so we don't get more bad data (I just repaired all the references on the live site!).

Do you think you can reproduce the error in a test?

chuenlok, line 13: I will make a separate test on wagn

chuenlok, line 13: https://github.com/wagn/wagn/pull/585
a test with a fix :)

PR #350, opened 2016-03-10
Metric value type

chuenlok:
1. a script to normalize values
2. updated the UI to edit value type
3. added and updated test cases for new changes
4. card changes import
5. show number to human in short format like `1.5K`
6. update the sorting to handle number


mod/wikirate/set/type_plus_right/metric/value_type.rb
```diff
@@ -0,0 +1,54 @@
+
+def all_numeric? metric_values
+  metric_values.each do |_key, values|
+    values.each do |mv|
+      return false unless number?(mv['value'])
+    end
+  end
+  true
```
xithan, line 8: What about 

``` ruby
metric_values.all? do |_key, values|
  values.all? { |mv| number? mv['value'] }
end
```

?

spec/spec_helper.rb
```diff
@@ -10,6 +10,24 @@
   end
 end
 
+def get_subcards_of_metric_value metric, company, content, year, source
```
xithan, line 13: Looks like this method is more about 'creating/initializing' than 'getting'.
Anyway, I think you don't need that. There is an API for creating metrics with values and in the branch I'm working on it's now very powerful and much better documented than it used to be: https://github.com/xithan/wikirate/blob/all_metrics/mod/01_rating/lib/card/metric.rb#L34

mod/wikirate/set/type_plus_right/metric/value_type.rb
xithan: or even shorter and equally fast (but much more confusing because of the negations; I favor the other solution):
`!metric_values.any? {  |_key, values| !values.all? { |mv| number? mv['value'] } }`

db/migrate_cards/data/new_metric_value_type.json
```diff
@@ -0,0 +1,77 @@
...
+      {
+        "name": "Metric+value type+*type plus right+*input",
+        "type": "Pointer",
+        "content": "[[select]]"
+      },
+      {
+        "name": "Metric+value type+*type plus right+*options",
+        "type": "Pointer",
+        "content": "[[Number]]\n[[Monetary]]\n[[Category]]\n[[Free Text]]"
+      },
```
ethn, line 33: need codenames for these.

db/migrate_cards/data/new_metric_value_type.json
```diff
@@ -0,0 +1,77 @@
...
+      {
+        "name": "Metric+value type+*type plus right+*help",
+        "type": "Basic",
+        "content": "<p>Value Type <span>is more important in computing than anywhere else in technology because software is so complicated. Beauty is the ultimate defence against complexity.</span></p>"
+      },
```
ethn, line 38: !

mod/01_rating/set/type/metric.rb
```diff
@@ -16,6 +17,64 @@
   def view_caching?
     true
   end
+
+  view :value_type_edit_modal_link do |args|
```
ethn, line 21: this should probably be in right/value_type  (or type_plus_right/metric/value_type, but the first one is probably fine)

ethn, line 21: if we're only changing args, this can be in an args method rather than a view.  (eg default_modal_link_args)

mod/01_rating/set/type/metric.rb
```diff
@@ -16,6 +17,64 @@
...
+    detail_card = Card.fetch "#{card.name}+#{subcard_name}"
+    subformat(detail_card).render_content
+  end
+
+  def default_edit_args args
+    edit_args args
+    super(args)
+  end
+
+  def edit_args args
```
ethn, line 66: why is this a separate method?

mod/01_rating/set/type/metric.rb
```diff
@@ -16,6 +17,64 @@
...
+      },
+      html_args: {
+        class: 'btn btn-default slotter'
+      },
+      text: text
+    }
+    render_modal_link(args.merge(edit_args))
+  end
+
+  view :short_view do |_args|
```
ethn, line 45: this should probably be called something like "detail_view" and should be in right/value_type

ethn, line 45: perhaps a separate method for determining the detail card?

mod/01_rating/set/type/metric.rb
```diff
@@ -16,6 +17,64 @@
...
+    return unless args[:edit_value_type]
+    args[:structure] = 'metric value type edit structure'
+  end
+
+  def edit_slot args
+    if args[:edit_value_type]
+      super args.merge(core_edit: true)
+    else
+      super args
+    end
```
ethn, line 76: marginal improvement?

```
super(args[:edit_value_type] ? args.merge(core_edit: true) : args)
```

mod/01_rating/set/type/metric_value.rb
```diff
@@ -48,20 +48,50 @@ def subfield_exist? field_name
...
   end
 end
 
+def number? str
+  true if Float(str)
+rescue
+  false
+end
+
+event :validate_value_type, :validate, on: :save do
```
ethn, line 71: probably want separate methods for each validation.

mod/01_rating/set/type/metric_value.rb
```diff
@@ -190,36 +223,62 @@ def legend args
     subformat(card.metric_card)._render_legend args
   end
 
+  def currency
+    return unless (value_type = Card["#{card.metric_card.name}+value type"])
```
ethn: again, if this is in +value_type, we can skip rechecking that we have a value type card.

mod/01_rating/set/type/metric.rb
```diff
@@ -1,9 +1,10 @@
-card_accessor :vote_count, :type=>:number, :default=>"0"
-card_accessor :upvote_count, :type=>:number, :default=>"0"
-card_accessor :downvote_count, :type=>:number, :default=>"0"
+card_accessor :vote_count, type: :number, default: '0'
+card_accessor :upvote_count, type: :number, default: '0'
+card_accessor :downvote_count, type: :number, default: '0'
```
ethn, line 3: add value_type accessor

mod/01_rating/set/type/metric_value.rb
```diff
@@ -190,36 +223,62 @@ def legend args
...
+
   view :concise do |args|
-    %{
+    %(
       <span class="metric-year">
         #{card.year} =
       </span>
+      <span class="metric-unit">
+        #{currency}
+      </span>
```
ethn, line 240: This file really  shouldn't know anything about value types.  Could do something like:
`subformat(value_type_card).render_concise_details`

...and then have different detail handling for different metric types.  (?)

mod/01_rating/set/type/metric_value.rb
ethn, line 71: and...again, too specific for this file.

chuenlok:
Thanks for the comment! They are really helpful. I am going make a new set for metric value type like the source_type so that I can separate the specific code.

PR #351, opened 2016-03-12
Metric value import ui

chuenlok:
There are two duplicated files, `mod/wikirate/set/abstract/import.rb`, `mod/wikirate_source/set/abstract/import.rb`. I tried to `include Card::Set::Abstract::Import` in `mod/wikirate_source/set/type_plus_right/source/file/import.rb` but it is not working. I need to create `mod/wikirate_source/set/abstract/import.rb` in order to make it work.

Do I miss sth to include a abstract module from another mod?


ethn:
I'll have to solve the inclusion issue.  The duplication is unacceptable.

mod/wikirate/set/abstract/import.rb
```diff
@@ -0,0 +1,222 @@
+def metric_value_subcards metric, company, year, value, source
+  {
+    '+metric' => { 'content' => metric },
+    '+company' => { 'content' => company },
+    '+value' => { 'content' => value, :type_id => PhraseID },
+    '+year' => { 'content' => "[[#{year}]]", :type_id => PointerID },
```
ethn: these types should not be necessary; aren't they in default rules?

mod/wikirate/set/type/metric_value_import_file.rb
```diff
@@ -0,0 +1,187 @@
...
+  metric_value_card_name = [metric, company, year].join '+'
+  mv_subcards = metric_value_subcards metric, company, year, value, source
+  create_or_update_mv_card metric_value_card_name, mv_subcards
+end
+
+def handle_redirect
+  if errors.empty?
+    abort success: {
+      slot: {
+        success_msg: 'Import Successfully <br />'
```
ethn: br is ugly here.

mod/wikirate/set/abstract/import.rb
```diff
@@ -0,0 +1,222 @@
...
+end
+
+event :import_csv, :prepare_to_store,
+      on: :update,
+      when: proc { Env.params['is_metric_import_update'] == 'true' } do
+  import_csv_information
+end
+
+def clean_corrected_company_hash
+  hash = Env.params[:corrected_company_name]
```
ethn: hack, no?

mod/wikirate/set/abstract/import.rb
```diff
@@ -0,0 +1,222 @@
...
+            'content' => source, 'type_id' => PhraseID
+          }
+        }
+      }
+    }
+  }
+end
+
+def import_csv_information
+end
```
ethn: hack without comment.

mod/wikirate/set/abstract/import.rb
```diff
@@ -0,0 +1,222 @@
...
+  else
+    Card.create type_id: Card::MetricValueID, subcards: subcard
+  end
+end
+
+def csv_rows
+  # transcode to utf8 before CSV reads it.
+  # some users upload files in non utf8 encoding.
+  # The microsoft excel may not save a CSV file in utf8 encoding
+  CSV.read(file.path, encoding: 'windows-1251:utf-8')
```
ethn, line 122: how does this explain why windows-1251 is hard coded here?

mod/wikirate/set/abstract/import.rb
```diff
@@ -0,0 +1,222 @@
...
+end
+
+def clean_html? # return always true ;)
+  false
+end
+
+format :html do
+  def aliases_hash
+    aliases_hash = {}
```
ethn: seems like a case for `@aliases_hash ||= begin...`

mod/wikirate/set/type/metric_value_import_file.rb
```diff
@@ -0,0 +1,187 @@
...
+
+def create_metric_value_from_hash mv_hash, corrected_company_hash
+  metric, company, year, value, source = parse_hash mv_hash
+  correct_company_key = "#{metric}+#{company}+#{year}"
+  potential_alias_name = company
+  company = get_final_company_name correct_company_key, company,
+                                   corrected_company_hash
+  # add corrected name to alias
+  add_corrected_name_to_alias corrected_company_hash[correct_company_key],
+                              potential_alias_name
```
xithan: What do we want to do here? I would think that "corrected name" means the new (correct) name but what we do here is adding the old (wrong) name as alias. Looks like a bug to me but on the other hand `potential_alias_name` is explicitly set to the old name before we get the new name. So I'm not sure if it's just a naming issue.

xithan: after a closer look to the code I assume it's a naming issue. The corrected name is the one that exists in the database so I guess we want to add to its alias card the other name.

mod/wikirate/set/type/metric_value_import_file.rb
```diff
@@ -0,0 +1,187 @@
...
+end
+
+format :file do
+  include Type::File::FileFormat
+end
+
+event :validate_import, :prepare_to_validate,
+      on: :update,
+      when: proc { Env.params['is_metric_import_update'] == 'true' } do
+end
```
xithan: empty event?

mod/wikirate_source/set/type_plus_right/source/file/import.rb
```diff
@@ -1,237 +1,62 @@
...
+end
+
+def metric_year_content
+  metric_year_card.item_names.first
 end
 
-event :import_csv, :prepare_to_store,
+event :validate_import, :prepare_to_validate,
       on: :update,
       when: proc { Env.params['is_metric_import_update'] == 'true' } do
```
xithan, line 25: Why do we have this 'is_metric_import_update' flag? Do we update import files without intending to import something?

ethn, line 25: good question.

mod/wikirate/set/type/metric_value_import_file.rb
ethn: that sounds right.  need more code comments.

PR #355, opened 2016-03-17
New metric value

srivig:


mod/wikirate_source/set/type/source/source.rb
```diff
@@ -67,6 +67,27 @@ def analysis_cards
...
 
 format :html do
+  view :new do |args|
+    # return super(args)
+    if Env.params[:preview]
+      form_opts = args[:form_opts] ? args.delete(:form_opts) : {}
+      form_opts[:hidden] = args.delete(:hidden)
+      form_opts['main-success'] = 'REDIRECT'
+      form_opts['data-form-for'] = 'new_metric_value'
+      form_opts[:class] = "card-slot"
```
xithan, line 77: not necessary to repeat form_opts. Use merge! for example:

```
form_opts.merge! hidden: args.delete(:hidden),
                            'main-success' => ...
```

mod/01_rating/set/type/metric_value.rb
```diff
@@ -240,26 +238,27 @@ def legend args
...
-                          data:{
-                            toggle:'collapse-next',
-                            parent:'.value',
-                            collapse:'.metric-value-details'}
-                          )
+                                        :i,
+                                        '',
+                                        class: 'fa fa-caret-right
+                                                fa-lg margin-left-10
+                                                btn btn-default btn-sm',
```
xithan, line 256: I think this means you have a lot of whitespace and line breaks in your class definition.
I would do it like this:

```
'string1' \
'string2'
```

PR #357, opened 2016-03-21
metrics

xithan:
not ready yet


mod/01_rating/lib/card/metric.rb
```diff
@@ -0,0 +1,104 @@
...
+      @metric.create_value args
+    end
+
+    def method_missing company, *args
+      args.first.each_pair do |year, value|
+        create_value company, year, value
+      end
+    end
+
+    def add_values_to metric
```
ethn, line 28: not used?

mod/01_rating/lib/javascript/drag_and_drop.js.coffee
```diff
@@ -0,0 +1,260 @@
+sortDragItems = (list, key, order='desc') ->
+  order = if order == 'desc' then 1 else -1
+  $(list).find('.yinyang-row').sortElements (a, b) ->
```
ethn: probably obvious, but we should probably try to update the name "yinyang" at some point, since the yinyang design elements are all gone now...

mod/01_rating/set/right/variables.rb
```diff
@@ -0,0 +1,117 @@
+include Type::Pointer
+
+def metric_card
+  binding.pry
```
ethn: hound didn't mention binding.pry?

mod/01_rating/set/type/metric.rb
```diff
@@ -2,7 +2,176 @@
...
+def random_valued_company_card
+  return unless (rvc = random_value_card)
+  rvc.company_card
+end
+
+def metric_value_cards opts={}
+  Card.search({ left: { left: { left: name },
+                right: { type: 'year' } },
+                right: 'value'
+              }.merge(opts))
```
ethn: can we use the "Metric Value" type to make this simpler? (and faster)

ethn:
As usual, when I let myself get this far behind, it's tough to get my head around it all, but there's clearly lots of good work here – more than I can digest!

I'm going to have a quick look at semaphore and the changes already discussed on henry's code and see if I can get caught up to usefulness that way.

PR #364, opened 2016-04-05
metrics and metric value import

xithan:
combines #357 and #351


mod/01_rating/set/type_plus_right/metric/formula.rb
```diff
@@ -40,13 +40,21 @@ def translation_hash
   return {} unless content.present?
   JSON.parse(content)
 rescue JSON::ParserError => _e
-  fail Card::Error, 'fail to parse formula for categorical input'
+  content = '{}'
```
ethn, line 43: is there a reason for this that hound and I aren't seeing?

mod/wikirate_source/set/type_plus_right/source/file/import.rb
```diff
@@ -1,237 +1,54 @@
+include_set Abstract::Import
```
ethn, line 1: well this looks muuuuuch better :)

PR #371, opened 2016-04-11
Fix url iframable check

chuenlok:
`curl.head` is returning `nil`. 
However, in raw ruby console and test environment, it is returning correct http header string.

`curl.header_str` is working fine and is used to replace `curl.head`.

### DIFF
--- a/mod/wikirate_source/set/self/source.rb
+++ b/mod/wikirate_source/set/self/source.rb
@@ -12,12 +12,12 @@ format :json do
       metadata.error = 'empty url'
       return metadata.to_json
     end
-    begin      
+    begin  
       metadata.website = URI(url).host
-    rescue    
+    rescue
     end
     if !metadata.website
-      metadata.error = 'invalid url' 
+      metadata.error = 'invalid url'
       return metadata.to_json
     end
     duplicates = Source.find_duplicates url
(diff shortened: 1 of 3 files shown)


mod/wikirate_source/set/self/source.rb
```diff
@@ -39,108 +39,83 @@ def self.find_duplicates url
...
+  def iframable_options url
+    # escape space in url, eg,
+    # http://www.businessweek.com/articles/2014-10-30/
+    # tim-cook-im-proud-to-be-gay#r=most popular
+    url.gsub!(/ /, '%20')
+    curl = Curl::Easy.new(url)
+    curl.follow_location = true
+    curl.max_redirects = 5
+    curl.http_head
+    header_str = curl.header_str
```
chuenlok, line 66: instead of `curl.head`

mod/wikirate_source/set/source_type/wikirate_link.rb
```diff
@@ -106,12 +106,12 @@ def download_file_and_add_to_plus_file url
...
 end
 
-def get_curl url
+def get_header url
   curl = Curl::Easy.new(url)
   curl.follow_location = true
   curl.max_redirects = 5
   curl.http_head
-  curl
+  curl.header_str
```
chuenlok, line 114: instead of `curl.head`

PR #376, opened 2016-04-13
get source type set cards right

xithan:


db/migrate_cards/20160413193247_source_type_set_cards.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def create_source_type_set name
+    create_card!(
+      name: "#{name}+*source type",
+      type_id: Card::SetID,
+      content: '{"type":"source",' +
+               %("right_plus":["*source type", {"refer_to":"#{name}"}]})
+    )
+  end
+end
```
ethn, line 18: this still confuses me.

It seems like:
1. there are Source cards with +_source type cards that determine their type
2. this creates a rule where ALL "+_source type" cards are sets.

Doesn't #2 override #1 and break things?

xithan, line 18: this creates a rule where "File+*source type", "Link+*source type" and "Text+*source type" are sets so the answer is no, it doesn't override and break things. The other approaches I used before with "source type+*source type+*type plus right" and "*source type+*right" broke things.

PR #377, opened 2016-04-13
Demofixes

ethn:
trying to get demo tests passing.

_Most_ of what I've fixed so far is just test artifacts, but there may be a real bug fix or two in here.


ethn:
doh.  made PR to master.  fixing...

PR #382, opened 2016-04-15
Fixing tests

chuenlok:


mod/caching/spec/set/type_plus_right/wikirate_company/source_spec.rb
```diff
@@ -5,7 +5,7 @@
     samsung_sources = Card.fetch "Samsung+source"
     expect(samsung_sources.cached_count).to eq 0
     source = get_a_sample_source
-    company_list = source.fetch trait: :wikirate_company
+    company_list = source.fetch trait: :wikirate_company, new: {}
```
chuenlok, line 8: I have `Company+*right` in the test db but the fetch is not returning me a virtual card. Do I miss sth?

ethn, line 8: Company+*right is a Set.  You would need Company+*right+*structure for it to be a virtual card.  But if it were a virtual card, then the next line (`company_list.add_item!...`) wouldn't make any sense, because you can't edit the content of a structured card.

PR #383, opened 2016-04-15
update test db and fix cuke test

chuenlok:


mod/wikirate/features/metric_create.feature
```diff
@@ -8,6 +8,6 @@ Feature: create metrics
     When I go to new metric
     And I choose "Researched"
     And I fill in "Metric Title" with "MyResearch"
-    And I fill in "Question" with "my question"
-    And I fill in "Value Type" with "Number"
+    And I fill in "card[subcards][+question][content]" with "my question"
```
chuenlok, line 11: textarea does not have the id linked to the label so I have to use it's name field.

ethn, line 11: Cucumber files should be things that Vishal can read and make sense of.  All this complexity should be hidden in steps. 

I'm going to merge this, but we need to clean up our dirty cukes.

PR #388, opened 2016-04-20
fix sum field for WikiRatings

xithan:


mod/01_rating/set/type_plus_right/metric/formula/wiki_rating.rb
```diff
@@ -13,13 +13,13 @@
         subformat(metric)._render_weight_row(args.merge(weight: weight))
       end
     end
-    sum_field =
+    sum =
       if table_content.empty?
         { content: sum_field, class: 'hidden' }
       else
         sum_field
```
xithan, line 20: obviously bad style but I wasn't aware that's it's not even working if you  assign the return value of a method to the method's name. The variable is nil after that.

ethn, line 20: I wouldn't have guessed that either.  good catch.

PR #389, opened 2016-04-20
Unknown metric value

chuenlok:
minimal update to fit `unknown` for metric value.


mod/wikirate/set/type_plus_right/metric/all_values.rb
```diff
@@ -63,15 +63,20 @@ def count _params={}
...
+    options[:remote] = true
     sort_by, sort_order = card.sort_params
     paging_args = @paging_path_args.merge(sort_by: sort_by,
                                           sort_order: sort_order)
     link_to raw(text), path(paging_args), options
   end
 
+  def unknown_value? value
+    value.casecmp('unknown') == 0
+  end
```
ethn, line 76: if this were a card method, we could reuse this code in the validation event.

PR #393, opened 2016-04-21
a few test cases and fixes

chuenlok:
1. alias is not added to input company in import if there is no matching company because the JSON is not correct.
2. source's website is not in pointer type
3. on source page, the page keeps going back to metric tab
4. metric+company's latest year cached is not correct

### DIFF
--- a/mod/wikirate/set/type_plus_right/metric/value_type.rb
+++ b/mod/wikirate/set/type_plus_right/metric/value_type.rb
@@ -2,7 +2,9 @@
 def all_numeric? metric_values
   metric_values.each do |_key, values|
     values.each do |mv|
-      return false unless number?(mv['value'])
+      if !number?(mv['value']) && mv['value'].casecmp('unknown') != 0
+        return false 
+      end
     end
   end
   true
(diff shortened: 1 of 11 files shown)


chuenlok:
1. fixed creating empty source titles
2. changing value type allows value `unknown`

mod/wikirate/set/type_plus_right/metric/value_type.rb
```diff
@@ -13,7 +15,10 @@ def all_values_in_options? metric_values, options_card
                                     context: :raw
   metric_values.each do |_key, values|
     values.each do |mv|
-      return false unless options.include?(mv['value'].downcase)
+      if !options.include?(mv['value'].downcase) &&
+         mv['value'].casecmp('unknown') != 0
```
ethn, line 19: too much cut and paste of this code.

mod/wikirate_source/spec/set/type_plus_right/source/file/import_spec.rb
```diff
@@ -28,6 +28,10 @@ def metric_answer_name company
     metric + '+' + company_name(company) + '+2015'
   end
 
+  def alias_card company
+    Card.fetch "#{company}+aliases", new: {}
+  end
```
ethn, line 33: this should be in type/company

PR #398, opened 2016-04-28
fetch update

xithan:
and tweaked fetch calls


xithan:
needs a test db update and for that an production export update is needed

ethn:
ok to close this?

ethn:
shall we close this?  or no... you probably want to keep it around until the fetch upgrade is in place, right?

PR #400, opened 2016-04-28
fix test db

chuenlok:
Updated `production_export` on live


mod/caching/spec/set/type_plus_right/wikirate_company/source_spec.rb
```diff
@@ -2,11 +2,11 @@
 
 describe Card::Set::TypePlusRight::WikirateCompany::Source do
   it 'updates cached count' do
-    samsung_sources = Card.fetch "Samsung+source"
+    samsung_sources = Card.fetch 'Apple Inc+source'
```
chuenlok, line 5: Samsung has a source by default now

mod/caching/spec/set/type_plus_right/wikirate_company/source_spec.rb
```diff
@@ -2,11 +2,11 @@
...
     expect(samsung_sources.cached_count).to eq 0
     source = get_a_sample_source
     company_list = source.fetch trait: :wikirate_company, new: {}
-    company_list.add_item! 'Samsung'
+    company_list.add_item! 'Apple Inc'
```
ethn, line 9: why did this matter?

PR #406, opened 2016-05-03
fix adding metric value from source

chuenlok:
1. fixed system level too deep problem while deleting metric
2. fixed metric import and tests, removed new_source handling for metric value
3. fixed metric value renaming


mod/wikirate/set/abstract/import.rb
```diff
@@ -5,24 +5,62 @@
...
+  {
+    '+*source_type' => { content: '[[Link]]' },
+    '+Link' => { content: url, type_id: PhraseID }
+  }
+end
+
+def create_source url
+  Env.params[:sourcebox] = 'true'
+  source_card = add_subcard '', type_id: SourceID, subcards: source_args(url)
+  source_card.director.catch_up_to_stage :prepare_to_store
```
chuenlok, line 53: I tried to `catch_up_to_stage :finalize` directly but the `abort :success` in the `sourcebox` handling is not stopping the card creation that creates duplicated sources if the card exists with the `url`.

The returned `source_card` is not the source card in db so I need to check if it exists before I can fire it to finalize phase.

mod/01_rating/set/type/metric_value.rb
```diff
@@ -197,47 +142,20 @@ def add_company source_name
...
+      if Card.exists? source_name
+        add_report_type source_name
+        add_company source_name
+      else
+        errors.add :source, "#{source_name} does not exist."
+      end
     end
+  elsif action == :create
+    errors.add :source, 'does not exist.'
   end
```
ethn, line 158: That's a lot shorter!!

PR #407, opened 2016-05-05
update score value column name, add checking on the values

chuenlok:


mod/01_rating/set/right/variables.rb
```diff
@@ -87,7 +87,10 @@ def default_edit_args args
   end
 
   def metric_list
-    Card.search(type_id: MetricID, limit: 0).map do |m|
+    metrics = Card.search(type_id: MetricID,
```
chuenlok: this refers to https://www.pivotaltracker.com/story/show/118517081 `make wikirating selector use only scores`. It will only shows score metrics now instead of all metrics.

mod/01_rating/set/type_plus_right/metric/formula.rb
```diff
@@ -154,14 +151,29 @@ def get_nest_defaults _nested_card
 
 event :validate_category_translation, :validate,
       when: proc { |c| c.translate_formula? } do
-  # TODO: Check if there is a translation for all value options
+  complete_translation_table.each do |value_pair|
```
chuenlok, line 154: add checking on the value added.

mod/01_rating/set/right/variables.rb
```diff
@@ -87,7 +87,15 @@ def default_edit_args args
...
   def metric_list
-    Card.search(type_id: MetricID, limit: 0).map do |m|
+    metrics =
+      if card.metric_card.metric_type_codename == :wiki_rating
+        Card.search(type_id: MetricID,
+                    right_plus: ['*metric type', { refer_to: 'Score' }],
+                    limit: 0)
+      else
+        Card.search(type_id: MetricID, limit: 0)
+      end
```
ethn, line 97: Could be tighter, eg:

```
wql = { type_id: MetricID, limit: 0 }
if card.metric_card.metric_type_codename == :wiki_rating
  wql[:right_plus] =  ['*metric type', { refer_to: 'Score' }]
end
Card.search(wql).map do |m|...
```

mod/01_rating/set/right/variables.rb
```diff
@@ -87,7 +87,15 @@ def default_edit_args args
   end
 
   def metric_list
```
ethn, line 89: I know this is old code, but I might consider a better name for this, like "variable_options".

mod/01_rating/set/type_plus_right/metric/formula.rb
ethn, line 154: does this break on "unknown"?

mod/01_rating/spec/lib/card/metric_spec.rb
```diff
@@ -81,6 +81,7 @@
 
     it 'creates score' do
       Card::Auth.as_bot do
+        researched_metrics
```
ethn, line 84: ??  comment?

mod/01_rating/set/type_plus_right/metric/formula.rb
ethn, line 154: (generally speaking, for now if a value is "unknown", we shouldn't score it.)

PR #409, opened 2016-05-09
improve error msg for import

chuenlok:
1. fixed a missing field in new metric value form
2. improved error msg for import
3. fixed a little error in `all_values`
4. updated some codes and added some comments based on the last PR's comment.

### DIFF
--- a/mod/01_rating/lib/javascript/script_metric_value.js.coffee
+++ b/mod/01_rating/lib/javascript/script_metric_value.js.coffee
@@ -53,7 +53,7 @@ $(document).ready ->
                                      metric)
       else
         source = $.urlParam('source')
-        if source != undefined
+        if source != null
           source = '&slot[source]=' + source
         else
           source = ''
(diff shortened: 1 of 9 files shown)


mod/01_rating/lib/javascript/script_metric_value.js.coffee
```diff
@@ -53,7 +53,7 @@ $(document).ready ->
                                      metric)
       else
         source = $.urlParam('source')
-        if source != undefined
```
chuenlok, line 56: $.urlParam is returning `null` instead of `undefined`.

mod/01_rating/set/type/metric_value.rb
```diff
@@ -345,26 +345,22 @@ def find_potential_sources company, metric
   end
 
   def set_hidden_args args
-    if !args[:source]
```
chuenlok, line 348: this was for old flow. While users click add new metric value from the source page, it will have source name in the args. It should still show the metric subcard field here.

PR #410, opened 2016-05-10
better error msg for missing value in options

chuenlok:
1. better error msg for missing value in options
2. fix inconsistent number format, show full value < 1M(was requested by Richard long time before)
3. allow comment in import

### DIFF
--- a/mod/01_rating/set/type/metric/creation.rb
+++ b/mod/01_rating/set/type/metric/creation.rb
@@ -66,6 +66,9 @@ def create_value_args args
       type_id: (args[:value].is_a?(Integer) ? NumberID : PhraseID)
     }
   }
+  if args[:comment].present?
+    create_args['+discussion'] = { comment: args[:comment] }
+  end
   add_value_source_args create_args, args[:source]
   create_args
 end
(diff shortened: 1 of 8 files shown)


mod/wikirate/set/type_plus_right/metric/value_type.rb
```diff
@@ -11,27 +11,34 @@ def invalid_metric_values metric_values
...
+def show_category_option_errors options_card, invalid_options
   url = "/#{options_card.cardname.url_key}?view=edit"
   anchor =
     <<-HTML
       <a href='#{url}' target="_blank">add the values to options card</a>
     HTML
-  errors.add :value, "Please #{anchor} first"
+  values =
+    if !invalid_options.nil?
+      "\"#{invalid_options.uniq.join('\", \"')}\" missing in the options."
```
ethn: how about " is not an option for this Metric"  ?

mod/01_rating/set/type/metric/creation.rb
```diff
@@ -66,6 +66,9 @@ def create_value_args args
       type_id: (args[:value].is_a?(Integer) ? NumberID : PhraseID)
     }
   }
+  if args[:comment].present?
+    create_args['+discussion'] = { comment: args[:comment] }
```
ethn, line 70: discussion has codename, right?

PR #411, opened 2016-05-11
Industry filter

ethn:
simple working version of filtering by company name/industry/project


mod/01_rating/set/right/metric_value_filter.rb
```diff
@@ -0,0 +1,142 @@
...
+  def filter_active?
+    Env.params.keys.any? { |key| filter_categories.include? key }
+  end
+
+  def default_core_args args
+    args[:buttons] = [
+      card_link(card.left, path_opts: { view: :content_left_col },
+                           text: 'Reset', class: 'slotter btn btn-default',
+                           remote: true),
+      button_tag('Filter', situation: 'primary', disable_with: 'Filtering')
```
ethn, line 46: "Filter" should probably  be ajax, too.  Feels like a little lost context when loading takes you back to the top of the page.

PR #414, opened 2016-05-12
make industry filtering use ajax

ethn:
filter won't work unless we use ajax, because the whole page is in the view cache.


mod/01_rating/set/right/metric_value_filter.rb
```diff
@@ -58,7 +58,7 @@ def default_core_args args
       </div>
       <div id="collapseFilter" class="panel-collapse collapse #{'in' if filter_active?}">
 
-        <form action="/#{action}" method="GET">
+        <form action="/#{action}?view=content_left_col" method="GET" data-remote="true" class="slotter">
```
ethn, line 61: could make a slot_form method for this kind of thing...

PR #416, opened 2016-05-13
fix over cache refreshing

chuenlok:


mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
```diff
@@ -1,33 +1,47 @@
...
+def calculate_count
+  (related_topic_from_source_or_note + related_topic_from_metric).uniq.size
+end
 
 # recount topics associated with a company whenever <source>+company is edited
 ensure_set { TypePlusRight::Source::WikiRateCompany }
 recount_trigger TypePlusRight::Source::WikirateCompany do |changed_card|
-  changed_card.item_names.map do |company_name|
+  names = Card::CachedCount.pointer_card_changed_card_names(changed_card)
+  names.map do |company_name|
```
ethn, line 35: I like the idea here.

Wouldn't it make more sense to put the change detection code in type/pointer and call it with something like "changed_card.item_names_updated_in_last_action"?

ethn, line 35: (same concept, just a question of where the code belongs)

PR #428, opened 2016-05-17
~0 issue

xithan:
show the first 3 non-negative digits after the decimal instead of ~0

### DIFF
--- a/mod/01_rating/set/type/metric_value/views.rb
+++ b/mod/01_rating/set/type/metric_value/views.rb
@@ -204,31 +204,33 @@ format :html do
     end
   end
 
-  def number_to_human_args
-    {
-      units: {
-        unit: '', billion: 'B', million: 'M', quadrillion: 'P', thousand: 'K',
-        trillion: 'T'
-      },
-      format: '%n%u',
-      delimiter: '',
-      precision: 3
-    }
+  def humanized_big_number number
+    number_to_human number,
+                    units: {
+                      unit: '', billion: 'B', million: 'M', quadrillion: 'P',
+                      thousand: 'K', trillion: 'T'
+                    },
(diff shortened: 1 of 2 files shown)


mod/01_rating/set/type/metric_value/views.rb
```diff
@@ -204,31 +204,33 @@ def grade
...
-      precision: (bigger_precision ? 3 : 1) }
+  def humanized_small_number number
+    less_than_one = number < 1
+    humanized = number_with_precision number,
+                                      delimiter: ',',
+                                      strip_insignificant_zeros: true,
+                                      precision: (less_than_one ? 3 : 1),
+                                      significant: less_than_one
+    (humanized == '0' && number > 0) ? '~0' : humanized
   end
```
ethn, line 226: very nice -- looks so much better now.

mod/01_rating/set/type/metric_value/views.rb
```diff
@@ -204,31 +204,33 @@ def grade
...
-    { delimiter: ',', strip_insignificant_zeros: true,
-      precision: (bigger_precision ? 3 : 1) }
+  def humanized_small_number number
+    less_than_one = number < 1
+    humanized = number_with_precision number,
+                                      delimiter: ',',
+                                      strip_insignificant_zeros: true,
+                                      precision: (less_than_one ? 3 : 1),
+                                      significant: less_than_one
+    (humanized == '0' && number > 0) ? '~0' : humanized
```
ethn, line 225: not sure I get the case when ~0 happens.

xithan, line 225: me too. left it there just in case I missed a case when it happens

PR #433, opened 2016-05-20
import msg and all values cache

chuenlok:
1. use company id instead of name in all values cache
2. import won't block metric values in the file with same value but showing warning msg after import

A script needs to be run after deployment to update all_values cache to replace company name with id. I did not put it in a migration because the script takes a lot of time to run that may block the db too long.


mod/01_rating/set/type/metric/creation.rb
```diff
@@ -32,17 +32,31 @@ def add_value_source_args args, source
...
+      error_msg.push("missing field(s) #{missing.join(',')}")
+      nil
+    end
+  end
+end
+
+def check_value_card_exist args, error_msg
+  return unless (value_name = extract_metric_value_name(args, error_msg))
+  if (value_card = Card[value_name.to_name.field(:value)])
+    unless value_card.content.casecmp(args[:value]) == 0
```
ethn, line 50: inconsistent style here.  `return unless`, `if`, and `unless` are all similar.  would read more cleanly if it were something like:

```
return unless (value_name = extract_metric_value_name args, error_msg)
return unless (value_card = Card[value_name.to_name.field :value])
return if value_card.content.casecmp(args[:value]) == 0
...
```

mod/wikirate/set/abstract/import.rb
```diff
@@ -37,14 +49,28 @@ def construct_value_args args
   create_args
 end
 
+def check_duplication_with_existing metric_value_name, source_card
+  slot_args = success.slot
+  if (source = Card[metric_value_name.to_name.field(:source)])
+    if source.item_cards[0].key == source_card.key
```
ethn, line 55: pls add comment to explain the difference between "identical" and "duplicate"

mod/caching/set/type_plus_right/metric/all_values.rb
```diff
@@ -22,7 +22,7 @@
 def calculate_count
   result = {}
   item_cards(default_query: true).each do |value_card|
-    company = value_card.company
+    company = value_card.company_card.id
```
ethn, line 25: company_id would be a clearer variable name, no?

mod/wikirate/set/abstract/import.rb
```diff
@@ -37,14 +49,28 @@ def construct_value_args args
...
 # @return updated or created metric value card object
 def parse_metric_value import_data, source_map
   args = process_metric_value_data import_data
   process_source args, source_map
   ensure_company_exists args[:company]
   return unless valid_value_data? args
   return unless (create_args = construct_value_args args)
-  check_duplication create_args[:name], args[:row]
+  check_duplication_in_subcards create_args[:name], args[:row]
+  return if check_duplication_with_existing create_args[:name], args[:source]
```
ethn, line 73: these two have the same naming pattern, but one returns and one doesn't.  pls comment and explain

PR #434, opened 2016-05-23
fix fetch fail while lvy is number

chuenlok:
fetch fails to get correct card here if `lvy` is a number but not string


ethn:
looks good, thanks.

PR #435, opened 2016-05-31
Tests

chuenlok:
1. cuke tests for metric value import and new metric value
2. Vignesh unified the source item view


srivig:
1. Organised css classes and removed unused css classes from *css 
2. Only metric designer or admin can add value on 'designer assessed' metrics
3. Changed the structure of browse company/topic page items to show only metric,topic / metric, company

db/migrate_cards/data/cards/metric+topic_count+*type_plu_right+*structure
```diff
@@ -0,0 +1,4 @@
+{ 
+  "referred_to_by": "_l+topics",
+  "return" : "count"
+}
```
ethn, line 4: question: do we have a pattern of creating count cards like this?

another approach would be to make it so that (metric)+topic returned the topic search; then you could use (metric)+topic+*count to return the count.

mod/wikirate/set/type_plus_right/metric/all_values.rb
```diff
@@ -107,7 +107,11 @@ def filter_by_industry wql, industry
 
 def get_cached_values
   cached_json = fetch(trait: :cached_count, new: {}).format.render_raw
-  JSON.parse(cached_json).with_indifferent_access || {}
+  cached_hash = JSON.parse(cached_json).with_indifferent_access || {}
+  cached_hash.keys.each do |key|
+    cached_hash[Card[key.to_i].name] = cached_hash.delete key
+  end
+  cached_hash
```
ethn, line 114: shouldn't this just be a new hash?

chuenlok, line 114: do you mean to create a new hash instead of adding new things to the hash parsed?

PR #437, opened 2016-06-13
Company filter

chuenlok:
I was trying to use wql to sort company by `company+metric+*cached count` and `company+topic+*cached count` but it seems wagn only takes the first joined table as sorting order. I did a hack to order by the second joined table. Please let me know if there is a right way to do that.

`mod/wikirate/set/right/filter_search_company.rb`


db/migrate_cards/20160519140218_script_wikirate_common.rb
```diff
@@ -0,0 +1,9 @@
+# -*- encoding : utf-8 -*-
+
+class ScriptWikirateCommon < Card::Migration
+  def up
+    create_or_update name: 'script wikirate common',
+                     type_id: Card::CoffeeScriptID,
```
ethn, line 6: please follow hound suggestions unless there's a good reason not to.

PR #438, opened 2016-06-14
Company page filter

chuenlok:
I was trying to use wql to sort company by company+metric+_cached count and company+topic+_cached count but it seems wagn only takes the first joined table as sorting order. I did a hack to order by the second joined table. Please let me know if there is a right way to do that.

mod/wikirate/set/right/filter_search_company.rb


mod/wikirate/set/right/filter_search_company.rb
```diff
@@ -0,0 +1,79 @@
+include_set Abstract::Filter
+
+def item_cards params={}
+  s = query(params)
+  raise("OH NO.. no limit") unless s[:limit]
+  query = Query.new(s, comment)
+  # sort table alias always stick to the first table, but I need the next table
+  sort = query.mods[:sort].scan(/c([\d+]).db_content/).last.first.to_i + 1
+  query.mods[:sort] = "c#{sort}.db_content"
```
chuenlok, line 9: hacky way to make the sort table correct

PR #439, opened 2016-06-15
prevent import from blocking the site

chuenlok:
I guess if metric values are created using standard stage way, all things will be bonded in one transaction which will block the tables until all metric values are created.


chuenlok:
it is not working.

PR #442, opened 2016-06-29
Filter

chuenlok:
I was trying to use wql to sort company by company+metric+_cached count and company+topic+_cached count but it seems wagn only takes the first joined table as sorting order. I did a hack to order by the second joined table. Please let me know if there is a right way to do that.

`mod/01_filter/set/abstract/filter.rb`


db/migrate_cards/20160621161913_correct_project_codename.rb
```diff
@@ -0,0 +1,9 @@
+# -*- encoding : utf-8 -*-
+
+class CorrectProjectCodename < Card::Migration
+  def up
+    project_card = Card[:campaign]
+    project_card.codename = "project"
+    project_card.save!
```
xithan: = `Card[:campaign].update_attributes! codename: "project"`

mod/00_utility/set/abstract/utility.rb
```diff
@@ -0,0 +1,8 @@
+
+def params_to_hash params
+  params.each_with_object({}) do |param, hash|
+    if (val = Env.params[param])
+      hash[param.to_sym] = val
+    end
+  end
```
xithan: = `Env.params.select { |key, val| val && params.include?(key)}.symbolize_keys`

Maybe `fetch_params` or `select_params` describes better what that method does

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+
+def raw_content
+  %(
+    {
+      "name":"dummy"
+    }
+  )
+end
+
+def params_to_hash params
```
xithan: duplicate

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+    if (val = Env.params[param])
+      hash[param.to_sym] = val
+    end
+  end
+end
+
+def search_wql opts, return_param=nil
+  wql = { type_id: WikirateCompanyID }
+  wql[:return] = return_param if return_param
+  filter_by_name wql, opts[:company] if opts[:company].present?
```
xithan: present? check is already in `filter_by_name`

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+def virtual?
+  true
+end
+
+def raw_content
+  %(
+    {
+      "name":"dummy"
+    }
+  )
```
xithan: a waste of lines for that short string

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+        No result
+      </div>
+    )
+  end
+
+  view :filter_form do |args|
+    formgroups = args[:formgroups] || [:name_formgroup]
+    html = formgroups.map do |fg|
+      optional_render(fg, args)
+    end
```
xithan: I think the style guide recommends {} for one-line blocks (if the line is not too long)

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+    select_filter :wikirate_topic, "Topic"
+  end
+
+  view :metric_formgroup do
+    select_filter :metric
+  end
+
+  view :company_formgroup do
+    select_filter :wikirate_company, "Company"
+  end
```
xithan, line 208: it's almost worth to generate these views, almost

mod/01_filter/set/right/filter_search_metric.rb
```diff
@@ -0,0 +1,96 @@
...
+def search_wql opts, return_param=nil
+  wql = { type_id: MetricID }
+  wql[:return] = return_param if return_param
+  filter_by_name wql, opts[:metric]
+  filter_by_designer wql, opts[:designer]
+  filter_by_topic wql, opts[:wikirate_topic]
+  filter_by_year wql, opts[:year]
+  filter_by_project wql, opts[:project]
+  wql
+end
```
xithan, line 17: They look all similar these `search_wql` methods.
Use `params_keys.each { |key| send("filter_by_#{key}", wql, opts[key] }` and move it to Abstract::Filter?

mod/00_utility/set/abstract/utility.rb
ethn: ...and the argument should be "keys"

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+  super(params)
+end
+
+def item_cards params={}
+  s = query(params)
+  raise("OH NO.. no limit") unless s[:limit]
+  query = Query.new(s, comment)
+  shift_sort_table query
+  query.run
+end
```
ethn, line 40: if these are overriding the defaults, we need comments to say why and how.

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+  def page_link text, page, _current=false, options={}
+    @paging_path_args[:offset] = page * @paging_limit
+    filter_args = {}
+    page_link_params.each do |key|
+      filter_args[key] = params[key] if params[key].present?
+    end
+    options[:class] = "card-paging-link slotter"
+    options[:remote] = true
+    link_to raw(text), path(@paging_path_args.merge(filter_args)), options
+  end
```
ethn, line 139: as noted in the other PR, seems very odd to have this here.  definitely need a comment.

mod/01_filter/set/abstract/filter.rb
```diff
@@ -0,0 +1,232 @@
...
+  view :project_formgroup do
+    select_filter :project
+  end
+
+  view :year_formgroup do
+    select_filter :year
+  end
+
+  view :topic_formgroup do
+    select_filter :wikirate_topic, "Topic"
```
ethn: shouldn't the label just be the cardname?

mod/01_filter/set/abstract/filter.rb
chuenlok, line 40: just added comment but it is not showing here...

mod/01_filter/set/abstract/filter.rb
chuenlok, line 139: same, just added comment but it is not showing here...

mod/01_filter/set/abstract/filter.rb
```diff
@@ -53,27 +60,17 @@ def virtual?
...
+  wql = { type_id: type_id }
   wql[:return] = return_param if return_param
-  filter_by_name wql, opts[:company] if opts[:company].present?
-  filter_by_industry wql, opts[:industry] if opts[:industry].present?
-  filter_by_project wql, opts[:project] if opts[:project].present?
+  params_keys.each do |key|
+    # link_to in #page_link with name will override the path
+    method_name = key.include?("_name") ? "name" : key
+    send("filter_by_#{method_name}", wql, opts[key])
+  end
```
xithan, line 73: 👍

mod/01_filter/set/abstract/filter.rb
chuenlok, line 208: I thought these can be overridden easily later if necessary. We can only define the views we need in method #default_filter_form_args instead of calling the rendering everywhere. I can change it if you have a better idea.

PR #460, opened 2016-08-04
Company metric cache update

chuenlok:


mod/01_filter/set/abstract/filter.rb
```diff
@@ -315,8 +296,8 @@ def checkbox_formgroup title, options, default=nil
   end
 
   view :importance_formgroup do
-    options = ["Not Voted", "Upvoted", "Downvoted"]
-    checkbox_formgroup "My Vote", options, ["upvoted", "not voted"]
+    options = ["I voted FOR", "I voted AGAINST", "I did NOT vote"]
+    checkbox_formgroup "My Vote", options, ["i voted for", "i did not vote"]
```
ethn, line 300: fwiw, it's ok if the hidden names are more technical.  in fact, they may be easier for developers to understand.  but not important to change here.

mod/01_filter/set/right/filter_search.rb
```diff
@@ -15,7 +15,11 @@ def params_keys
   []
 end
```
ethn, line 16: this is where I get confused.  usually params_keys = default + advanced, right?  Why is it different here?

mod/wikirate/set/type_plus_right/wikirate_company/all_metric_values.rb
```diff
@@ -114,17 +114,22 @@ def filter_by_type metric
...
       Env.params["value"]
     else
       "exists"
     end
-  return values.empty? if value == "none"
-  return !values.empty? if value == "exists"
-  time_diff = second_by_unit value
+  return values.empty? if unit == "none"
+  return !values.empty? if unit == "exists"
+  within_recent? unit, values
```
ethn, line 125: this looks to me like a good place to use `case`.

```
case unit
when 'none' then values.empty?
when 'exists' then values.any?
else ...
```

mod/wikirate/set/type_plus_right/wikirate_company/all_metric_values.rb
```diff
@@ -114,17 +114,22 @@ def filter_by_type metric
 end
 
 def filter_by_value values
-  value =
+  unit =
     if Env.params["value"].present?
       Env.params["value"]
     else
       "exists"
     end
```
ethn, line 122: lot of space for simple concept.  maybe:

```
pval = Env.params["value"]
unit = pval.present? ? pval : "exists"
```

mod/01_filter/set/abstract/filter.rb
```diff
@@ -141,17 +142,21 @@ def default_button_formgroup_args args
     args[:buttons] = buttons.join
   end
 
-  view :filter_form do |args|
-    formgroups = args[:formgroups] || [:name_formgroup]
+  def advance_formgroups args
```
ethn, line 145: technically this should probably be "advanced", too.

(sorry to whine about something so minor, but I keep reading it as a verb, and that gives it a very different meaning)

PR #463, opened 2016-08-16
override fixture:load to seed with dumped test db, update test db

chuenlok:


lib/tasks/db.rake
```diff
@@ -0,0 +1,16 @@
...
+# reseeding from the live may take too much time and
+# some data only exists in development, which will cause
+# insufficient data in the reseed
+Rake::Task["db:fixtures:load"].clear
+namespace :db do
+  namespace :fixtures do
+    desc "Load fixtures into the current environment's database. "\
+         "Load specific fixtures using FIXTURES=x,y"
+    task load: :environment do
+      system "env RAILS_ENV=test bundle exec rake wikirate:test:seed"
```
xithan, line 13: Doesn't work `Rake::Task["wikriate:test:seed"].invoke` here? I don't think we need the rails env change for that task.

README.md
```diff
@@ -43,6 +43,23 @@ Run `rake wikirate:test:reseed_data` to prepare test database.
...
 
+After Updating Test DB (Optional)
+----
+As Wagn updated transaction handling, tables will be truncated after every cucumber test.
+There is rake task to re-import the test db after every test. It is ok to run the cucumebr tests after the reseed of data, but it would be very slow because of regenerating the *all+*script and *all+*style cards.
+
+These steps could help reduce the testing time.
+1. Start the server in test environment
+```ruby
+RAILS_ENV=test bundle exec wagn s
```
xithan, line 54: No need to start the server. Just a `Card[:all, :script].update_machine_output` in the reseed task before we dump would do it.

PR #464, opened 2016-08-17
try fixing the test with new transaction api

chuenlok:
1. some styles update
2. I need to define the `transaction_in_stage` after I called validate. I am not sure if this is a good/correct way to do. If I defined `transaction_in_stage` while adding subcards, the metric value card's errors are cleared/not attached to the parent source file card.


xithan:
included in #465

PR #465, opened 2016-08-23
green light at the end of the tunnel

xithan:


lib/tasks/wikirate/importer.rb
```diff
@@ -0,0 +1,107 @@
...
+      if card
+        puts "updating card #{name} "\
+               "#{card.update_attributes!(attr)}".light_blue
+      else
+        puts "creating card #{name} #{Card.create!(attr)}".yellow
+      end
+    rescue => e
+      puts "Error in #{name}\n#{e}".red
+    end
+  end
```
ethn, line 73: no urgent comments here, but long term:  since there's type-specific stuff, I was wondering whether we might want there to be a card method (eg #import) that we could override.

eg

```
def import attrib={}
  assign_attributes attrib
  save!
end
```

lib/tasks/wikirate/test.rake
```diff
@@ -1,17 +1,21 @@
...
 
-# require 'pry'
 namespace :wikirate do
   namespace :test do
-    db_path = File.join Wagn.root, "test", "seed.db"
-    testdb = ENV["DATABASE_NAME_TEST"] ||
-             ((t = Wagn.config.database_configuration["test"]) &&
-             t["database"])
+    full_dump_path = File.join Wagn.root, "test", "seed.db"
+    base_dump_path = File.join Wagn.root, "test", "base_seed.db"
```
ethn, line 7: base_dump is test db?

mod/wikirate/features/claim.feature
```diff
@@ -57,6 +57,7 @@ Feature: note
     And I fill in "sourcebox" with "http://google.com/?q=ymca"
     And I press "Submit"
     And I wait until ajax response
+    And I wait until ajax response
```
ethn, line 60: the double wait does it?

ethn, line 60: (if this is necessary, we may want a comment)

lib/tasks/wikirate/test.rake
xithan: base_dump is incomplete test db (with imports but w/o migrations and test data). If something breaks at the end of the seed generation you can start with that instead of importing everything again.

mod/wikirate/features/claim.feature
xithan: One wait for adding the source and one for submitting. Could be that it was just luck that it worked.

lib/tasks/wikirate/importer.rb
xithan: How would that work in the "Card.create" case when there is no card object?

PR #480, opened 2016-09-02
double check before deploying to production

xithan:


lib/capistrano/tasks/deploy.cap
```diff
@@ -17,6 +17,16 @@ end
...
+    next unless fetch(:stage) == :production
+    puts <<-EOF
+
+************************** WARNING ***************************
+            You are about to deploy to production
+**************************************************************
+
+    EOF
+    ask :answer, "Are you sure you want to update wikirate.org? Type 'forgive me, Yoda'"
+    exit unless fetch(:answer) == 'forgive me, Yoda'
```
ethn, line 29: Love it.

PR #481, opened 2016-09-02
Bug fixes

srivig:
Just a Minor HTML structure fix

### DIFF
--- a/mod/01_rating/lib/stylesheets/style_drag_and_drop.scss
+++ b/mod/01_rating/lib/stylesheets/style_drag_and_drop.scss
@@ -332,8 +332,11 @@ $bright-text-color: #e5e5e5;
       color:$dark-text-color;
 
     }
-    >.data {
+    >div>.data {
       width: auto!important;
+    }
+    div>.header {
+        width: 73%;
     }
   }
   .value-item {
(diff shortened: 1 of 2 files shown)


mod/01_rating/lib/stylesheets/style_drag_and_drop.scss
```diff
@@ -332,9 +332,15 @@ $bright-text-color: #e5e5e5;
       color:$dark-text-color;
 
     }
-    >.data {
-      width: auto!important;
+    >div{
+      >.data {
+        width: auto!important;
```
ethn, line 337: is this the convention? no space before the bang?

srivig: 'bang' you mean '!important' ? ( it is optional to have space before that, as many css minification tools strips the whitespace )

ethn: essentially, yes.  "bang" is a short way to say "exclamation point".

my question was more about stylistic convention than functionality; I know that it works, but I don't know whether we'd generally prefer to have a space there (before minifying)

PR #484, opened 2016-09-06
revert sort select to non chosen style

chuenlok:
https://www.pivotaltracker.com/story/show/129425217


ethn:
our filtering tests must be pretty weak, huh?  Few if any of these filtering PRs seem to have any tests associated.  Do you think we could add some useful filtering tests this week?

chuenlok:
I will write some tests for the filtering. Don't merge this before I make it :)

PR #485, opened 2016-09-06
fix metric formula issues

xithan:


ethn:
nice. looks as minor as I hoped it was.

PR #496, opened 2016-09-16
Paging lost filter

chuenlok:


mod/wikirate/set/abstract/all_values.rb
```diff
@@ -141,27 +141,40 @@ def page_link_params
...
     paging_args
   end
 
+  def path args={}
+    # the filter name cause conflict with the path name
+    return super(args) unless args.delete(:replace_name)
+    if (name = args.delete(:name))
+      args[:_name_] = name
+    end
+    super(args).gsub("_name_=", "name=")
```
xithan, line 165: a better solution might be to improve `path` so that you can explicitly set params:
for example `path params: { name: ... }`

xithan:
I will merge it in my next PR to resolve the conflicts

PR #504, opened 2016-09-23
Pathmark

ethn:
use new path(mark: mark) handling.


xithan (changes requested):
What about the type args in account_links.rb?

ethn:
those are all :mark now:

```
  def new_cardtype_path opts
    return unless opts[:action] == :new
    opts.delete :action
    return unless opts[:mark]
    "new/#{path_mark opts}"
  end
```

PR #511, opened 2016-10-07
Huddle work

xithan:
- metric answer lookup table
  - needs testing
  - imported flag not implemented
  - `populate_metric_answers_lookup_table` is now a script instead of a migration 
  - not used 
- new layout
  - unfinished (parts of the old layout, that I moved recently from cards to code, is in wikirate_tabs.rb and wikirate_layout.rb. should be removed if the new one is ready)
- filter refactoring
  not finished, probably broken
- renames `rating` mod to `metrics` mod 
  (makes more sense to me. most of the code is about metrics. rating is only a type of metric)
- all topics cache
  - cache update methods not finished


xithan:
part of #520

PR #541, opened 2016-12-12
Varia

ethn:


mod/00_utility/set/abstract/wikirate_tabs.rb
```diff
@@ -34,43 +34,32 @@ def tab_list tabs
...
-                  </div>)
-    layout container: true, fluid: true, class: "yinyang nodblclick" do
-      row 6 do
-        col class: "border-right" do
-          row md: 12, xs: 12,
-              class: "clearfix #{type}-content #{type}-page-logo-container" do
-            col class: "nopadding left" do
-              link_to_card card.cardname.trunk, link_text,
-                           class: "inherit-anchor"
+    bs do
```
ethn, line 37: bugfix.  Topics were broken.

mod/01_metrics/spec/set/metric_type/formula_spec.rb
```diff
@@ -229,14 +234,16 @@ def update_formula new_formula
...
+      test_calculation [], "60.0"
+      test_calculation %w(Samsung 2015), "29.0"
+      test_calculation %w(Sony_Corporation), "9.0"
+      not_researched_card = calc_value_card "Death_Star", "1977"
+      expect(not_researched_card).to be_falsey
     end
   end
 
   it "handles wolfram formula" do
+    pending "restoring access to Wolfram API"
```
ethn, line 246: note: current behavior of pending is that test code is still executed, so FYI we're still making API calls.  Hmm.  Maybe I'll comment these out...

mod/wikirate/set/type/cardtype.rb
```diff
@@ -1,3 +1,9 @@
+format do
+  def default_header_args _args
+    voo.variant = :plural
+  end
+end
```
ethn, line 5: There has been a long-standing request to make Cardtype pages use plural in their headers.  (eg "Companies" on the Browse Companies page).  This was ridiculously easy.

mod/wikirate/set/type/wikirate_company/structure.rb
```diff
@@ -4,6 +4,10 @@
 card_accessor :wikipedia
 
 format :html do
+  def default_content_formgroup_args args
+    voo.edit_structure = [:image, :wikipedia]
+  end
```
ethn, line 9: when we delete these key structures, we need to provide some kind of edit view.

I think this makes basic sense, but it hasn't been designed/discussed.

mod/wikirate/set/type/cardtype.rb
xithan: 😄

PR #542, opened 2016-12-12
added support for converting IDs and codenames to cards/names

ethn:


mod/wikirate/spec/config/initializers/core_extensions_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe CoreExtensions do
+  context String do
+    it "tests whether a string represents a number" do
+      expect("6".number?).to eq(true)
+      expect("Yomama".number?).to eq(false)
+    end
+  end
+
+  context CoreExtensions::PersistentIdentifier do
+    it "converts into a cardname" do
```
xithan, line 10: subject doesn't seem right here. I miss something like `describe "#name"` but at least there are tests.

xithan:
why do we have this in wagn and wikirate?

mod/wikirate/spec/config/initializers/core_extensions_spec.rb
ethn: good point.  will fix.

ethn:
I started in WikiRate and then realized that was dumb.  it's out of wikirate now in the research group branch (though I forgot to move the tests.

PR #554, opened 2016-12-19
paging and charts

xithan:


mod/filter/set/abstract/00_filter_helper.rb
```diff
@@ -32,5 +29,9 @@ def paging_path_args args={}
 end
 
 format do
+  def limit
+    query_with_params[:limit]
+  end
+
```
ethn, line 35: is this different from default?

mod/filter/set/abstract/browse_filter_form.rb
```diff
@@ -45,19 +31,17 @@ def target_type_id
...
-def get_query params={}
-  search_args = filter_wql
-  add_sort_wql search_args, sort_param if sort?
-  params[:query] = search_args
-  super(params)
+def wql_hash
+  wql = filter_wql
+  wql[:limit] = 20
+  add_sort_wql wql, sort_param if sort?
+  wql
```
ethn, line 38: part of the reason I had "raw" in the name of this method was that I was thinking this method was generally *not* supposed to have limits and other things that are frequently varied in it.  maybe "core_query" would be better?

In any event, does it break our pattern to put the limit in here?

mod/metrics/lib/card/chart/range_chart.rb
```diff
@@ -0,0 +1,96 @@
...
+      def x_label_scale
+        { name: "x_label",
+          type: "ordinal",
+          range: "width",
+          domain: { data: "x_labels", field: "text", sort: true },
+          points: true }
+      end
+
+      def calculate_ranges
+        if bucket_size > 2
```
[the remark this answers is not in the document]
ethn: probably worth addressing this hound comment

ethn (approved):

mod/filter/set/abstract/browse_filter_form.rb
xithan: I got a lot of "no limit" exceptions im `item_cards` and `item_names` calls. That was the easiest way to fix that

PR #555, opened 2016-12-19
Companytopics

ethn:
tabs on company page, etc.


mod/caching/set/type_plus_right/wikirate_topic/wikirate_company.rb
```diff
@@ -1,28 +1,153 @@
...
+def count
+  relation.count
+end
 
-# # ... <source>+company is edited
-# ensure_set { TypePlusRight::Source::WikirateCompany }
-# recount_trigger TypePlusRight::Source::WikirateCompany do |changed_card|
+# returns array of ids
+def all
+  relation.all
```
xithan, line 65: does `all` work here as expected?

ethn: hmm...probably not.

PR #569, opened 2017-01-04
email update

xithan:


ethn (approved):
Loving that I can actually review card content differences now.

Looks much cleaner.

PR #572, opened 2017-01-05
tweak metric value form and fix source import

xithan:


mod/metrics/set/type/metric_value/editor.rb
```diff
@@ -16,7 +16,7 @@
   def prepare_nests_editor
     year = card.fetch trait: :year, new: { content: card.year }
     voo.editor = :nests
-    voo.edit_structure = [[:value, {}], [year, {}]]
+    voo.edit_structure = [[:value, {title: "Value"}], [year, {title: "Year"}]]
```
ethn, line 19: I think you can just use the short hand [:value, "Title"]

PR #575, opened 2017-01-06
Css fixes

srivig:


db/migrate_cards/data/cards/homepage_company_metric_item
```diff
@@ -1,25 +1,22 @@
-
+  
   <!--prototype: Company+MetricDesigner+MetricName+yinyang drag item -->
   <div class="metric-item value-item">
     <div>
+      <a class="inherit-anchor" href= "/{{_1|linkname}}?filter[metric]={{_llr+_lr|linkname}}">
```
ethn, line 5: I think we want to use a cgi_escape view (not linkname) inside the query section of urls.

mod/metrics/set/type/wikirate_company/new_answer_form.rb
```diff
@@ -1,14 +1,65 @@
+include_set Abstract::Table
+
 format :html do
   view :new_metric_value, cache: :never do
     frame do
-      haml_view :new_metric_value_form
+      output [_render_metric_side, _render_source_side]
```
ethn: haml_views are new. Are you confident this doesn't conflict with PK's work?

ethn (commented):
I'd like to be confident that the haml_view stuff doesn't mask a real conflict.

mod/metrics/set/type/wikirate_company/new_answer_form.rb
xithan: most of the green stuff looks like old code to me

db/migrate_cards/data/cards/homepage_company_metric_item
ethn: as in `{{_llr+_lr|cgi_escape_name}}`

we might want to alias that with something shorter, like `cgi_name`

db/migrate_cards/data/cards/homepage_metric_company_item
```diff
@@ -0,0 +1,19 @@
+<div class="yinyang-row">
+  <a href="/{{_l|linkname}}?filter[wikirate_company]={{_r|name}}">
```
ethn, line 2: again, cgi_escape_name here.

db/migrate_cards/data/cards/homepage_metric_company_item
```diff
@@ -0,0 +1,19 @@
+<div class="yinyang-row">
```
ethn, line 1: just a comment: we're trying to move away from yin-yang row.  We don't have to move quickly, but whenever you work in this area, try to make sure you're using it less and less.

db/migrate_cards/data/cards/homepage_metric_company_item
```diff
@@ -0,0 +1,19 @@
...
+  <a href="/{{_l|linkname}}?filter[wikirate_company]={{_r|name}}">
+  <div class="company-item value-item">
+    <div class="header no-hover">
+              <div class="logo">
+                      {{_r+image|core;size:small}} 
+              </div>
+          <div class="name">
+              {{_r|name}}
+          </div>
+    </div>
```
ethn, line 11: bad spacing here.

ethn (changes requested):
please address the cgi escape issue. then I think we're ready to merge.

PR #580, opened 2017-01-09
removed code from research page and weak attempt at better view cache…

ethn:
… clearing


mod/metrics/set/right/metric_value.rb
```diff
@@ -10,6 +10,7 @@ def fast_search_results
     Answer.fetch record_id: card.left.id
   end
 
+  # TODO - this should be a view on type/Metric_Record (which we don't yet have)
```
xithan, line 13: will be gone soon anyway (see comment at the top)

PR #583, opened 2017-01-10
the +companies view cache fix

xithan:
in case we need it


ethn:
Do you remember what the bug was that this fixed?  Wonder if our new code fixes it?

xithan:
no, I can't remember

xithan:
but I remember that we didn't merge it because you wanted to fix it with a view cache update

PR #584, opened 2017-01-10
refactor and fix record list

xithan:


db/migrate_cards/20161005120800_answer_lookup_table.rb
```diff
@@ -1,5 +1,6 @@
 class AnswerLookupTable < ActiveRecord::Migration
   def up
+    drop_table :answers
```
xithan, line 3: note to self: remove this

mod/metrics/set/ltype_rtype/metric/wikirate_company.rb
```diff
@@ -1,4 +1,6 @@
-include_set Abstract::WikirateTable
+include_set Abstract::WikirateTable # deprecated, but not clear if it is still used
```
ethn, line 1: Soonish I'd like to make it so that `Metric+Company+*ltype rtype+*default` gives us a Record (type), and then all of this is in Record.

Biggish migration, but tiny code change.

PR #593, opened 2017-01-26
flags and research page

xithan:


mod/metrics/spec/set/type_plus_right/metric_value/checked_by_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      Card::Env.params.delete "set_flag"
+    end
+
+    it "checks the metric value" do
+      expect(cb_card.item_names.size).to eq(1)
+      expect(cb_card.item_names).to include("Joe User")
+    end
+
+    let(:checked_by) do
+      metric_value_card.fetch trait: :checked_by
```
ethn: Are you now thinking to keep the "checked by" card, or is this an intermediate step towards  a representation that will support the double check requested status?

mod/metrics/set/type_plus_right/metric_value/value.rb
```diff
@@ -90,6 +90,11 @@ def scored_metric?
   end
 end
 
+event :update_double_check_flag, :validate, on: [:update, :delete] do
+  return unless left.fetch trait: :checked_by
+  attach_subcard cardname.left_name.field_name(:checked_by), content: ""
+end
```
ethn: there are times when our API makes fairly nuanced functionality concise ;)

mod/metrics/spec/set/type_plus_right/metric_value/checked_by_spec.rb
ethn: nevermind.  I see that you're supporting that functionality on this card.

mod/metrics/set/type_plus_right/metric_value/checked_by.rb
```diff
@@ -49,46 +85,76 @@ def checked_users
...
 
   def data_path
     card.cardname.url_key
   end
 
+  def check_button_text
+    text = "Double check"
+    return text unless card.check_requested?
+    text << " #{request_icon} requested by #{card.requester}"
+    text
```
ethn: or...
```
def check_button_text
  output ["Double check", check_button_request_credit]
end

def check_button_request_credit
  return unless card.check_requested?
  " #{request_icon} requested by #{card.requester}"
end
```

mod/metrics/set/type_plus_right/metric_value/checked_by.rb
```diff
@@ -1,46 +1,82 @@
...
   #   return _render_double_check_view
   # end
   view :missing do |args|
     if card.new_card? && card.left
-      Auth.as_bot do
-        card.save!
-      end
-      render(@denied_view, args)
+      Auth.as_bot { card.save! }
+      render @denied_view, args
```
ethn: wait, is this where those tens of thousands of blank checked_by cards are coming from?

Do we need this?  Couldn't we get away with a few `:unknown_ok` tags on a few views instead?

ethn: (I know you didn't add this.  Just sayin'.  While we're here...)

mod/metrics/view/self/research_page/source_side.haml
```diff
@@ -0,0 +1,11 @@
+- haml_wrap do
```
ethn, line 1: you deleted a haml_wrap method above.  is this still intended here?  (I notice there's also a haml_wrap in the wagn documentation...)

db/migrate_cards/20170124120401_research_page.rb
```diff
@@ -0,0 +1,31 @@
...
+  def up
+    ensure_card "Research Page", type_id: Card::SessionID,
+                codename: "research_page"
+
+    ensure_field_rules "Company"
+    ensure_field_rules "Metric"
+
+    ensure_card "Unknown", codename: "unknown"
+    ensure_card "Unknown+*right+*input", content: "checkbox"
+    ensure_card "Unknown+*right+*add help",
```
ethn: Perhaps this should be `*help` and we should add this field to the edit view (in case corrections are needed)

mod/metric_answer_lookup/set/type_plus_right/metric_value/value.rb
```diff
@@ -1,3 +1,5 @@
+card_accessor :unknown
```
ethn: the +Unknown card is a field of the +value and not the answer?  Why is that?

Also, why implement it as a pointer rather than a boolean?

mod/metrics/lib/card/fixed_company_answer_query.rb
```diff
@@ -28,7 +28,7 @@ def topic_query value
       metric_ids =
         Card.search right_plus: [Card::WikirateTopicID, { refer_to: value }],
                     return: :id
-      @restrict_to_ids[:metric_id] += metric_ids
```
ethn, line 31: I like this api change.

mod/metrics/set/ltype_rtype/metric/wikirate_company/record.rb
```diff
@@ -0,0 +1,118 @@
+format :html do
+  def all_answers
+    @result ||=
+      Answer.fetch({ record_id: card.id }, sort_by: :year, sort_order: :desc)
+  end
```
ethn: should this be a model method?

mod/metrics/set/type_plus_right/metric_value/checked_by.rb
ethn: It would be great if you could try to get rid of this `save!` call.  I know it's not originally your mess, but it's still adding junk data.

ethn: (I think)

ethn (approved):
I'm going to go ahead and merge this and deploy to staging.  But it would be great if you could review my comments. (Some of them are a week or so old, but I'm not sure you've seen them).

xithan:
hm, I see  comments in nine places but no one is older than an hour

mod/metrics/set/ltype_rtype/metric/wikirate_company/record.rb
xithan: I wasn't sure if `@result` gets cleared if it is a class instance variable on `Answer`

mod/metrics/view/self/research_page/source_side.haml
xithan: I moved it from wikirate it to wagn

mod/metric_answer_lookup/set/type_plus_right/metric_value/value.rb
xithan: I guess you mean toggle. I forgot that we have that.

db/migrate_cards/20170124120401_research_page.rb
xithan: Is there an existing example for that?  A field that appears in edit view but not in edit_in_form view?

mod/metric_answer_lookup/set/type_plus_right/metric_value/value.rb
xithan: It's a field of value because I consider it as an interface to change the value. It's not data that we use. Actually, unknown fields are not stored in the database.  And our api doesn't support to manipulate siblings very well. It's more straight forward that +value checks if there is a +unknown subfield than to check if a +unknown is part of the same act.

xithan: toggle cards have a very limited api. I have to override the editor view to get a label

ethn: thanks for that explanation.  let's add `#not stored in database` as a comment on this line.

PR #596, opened 2017-02-09
multi-category metrics

xithan:


mod/metric_answer_lookup/lib/answer.rb
```diff
@@ -1,7 +1,94 @@
...
+
+    def card_id cardish
+      case cardish
+      when Integer then
+        cardish
+      when Card then
+        cardish.id
+      end
+    end
+  end
```
ethn, line 89: should @card.to_i return an id?  would that obviate this?

mod/metrics/set/type_plus_right/metric_value/value.rb
```diff
@@ -8,7 +8,7 @@ def value_card
 end
 
 def value
-  content
+  item_names.join ", "
```
ethn, line 11: we run this for all values?  interesting.  could it break on multiline free-text?

ethn (approved):
good looking code

mod/metrics/set/type_plus_right/metric_value/value.rb
xithan: We don't have interface for multiline free-text. It's a phrase card. It would replace linebreaks with commas.

PR #601, opened 2017-02-13
Homepage design

srivig:


mod/wikirate/lib/javascript/test.js
```diff
@@ -0,0 +1,21 @@
...
+            console.log (cnt + "Liked");
+        }
+    }
+    else
+    {	console.log('Total HREF Links Processed: ' + anchorTag.length);
+        clearInterval(TimerFunCall);
+    }
+    i=i+1;
+    return i;
+}
```
xithan: what is this js file about?

srivig: I have no idea. I am deleting it

xithan: 😄

srivig: ahh, I just remember, I wrote that script to 'like' all the birthday wishes on my facebook wall. Looks like I saved it in WikiRate dev directory >.<

PR #618, opened 2017-03-01
profile subtabs

xithan:


db/migrate_cards/20170301085535_double_check_overhaul.rb
```diff
@@ -0,0 +1,23 @@
...
+    ensure_card "request", codename: "request"
+    ensure_card "check requested by+*right+*default", type_id: Card::PointerID
+    Card::Cache.reset_all
+
+    update_checked_by_cards
+  end
+
+  def update_checked_by_cards
+    Card.search(right: { codename: "checked_by" }).each do |card|
+      if card.item_names.first == "request"
```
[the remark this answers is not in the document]
ethn: (agree with :dog:)

mod/metrics/set/type/metric/creation.rb
```diff
@@ -6,6 +6,25 @@ def create_value_options options
...
+
+event :ensure_title, :prepare_to_store, on: :save, changed: :name do
+  return if Card.fetch_type_id(metric_title) == MetricTitleID
+  attach_subcard metric_title, type_id: MetricTitleID
+end
+
+def valid_designer?
+  Card.fetch_type_id(metric_designer).in? [ResearchGroupID, UserID,
+                                           WikirateCompanyID]
+end
```
ethn, line 28: yay for this.

mod/metrics/set/type/metric/creation.rb
```diff
@@ -6,6 +6,25 @@ def create_value_options options
...
+  if (card = Card[metric_designer])
+    errors.add :metric_designer, "invalid type #{card.type_name}"
+  else
+    attach_subcard metric_designer, type_id: ResearchGroupID
+  end
+end
+
+event :ensure_title, :prepare_to_store, on: :save, changed: :name do
+  return if Card.fetch_type_id(metric_title) == MetricTitleID
+  attach_subcard metric_title, type_id: MetricTitleID
```
ethn: would we catch it if someone tried to use a name that was already a non-metric-title card?

mod/metrics/set/type_plus_right/metric_value/checked_by.rb
```diff
@@ -19,19 +20,32 @@ def other_user_requested_check?
 end
 
 def checked?
-  !check_requested? && checkers.present?
+  checkers.present?
 end
 
 def check_requested?
-  items.first == "request" && items.size <= 2
+  items.first == "request"
```
ethn, line 37: confused.  it looks to me like you were moving away from this representation in the migration above.  aren't we representing this with `check_requested_by`?

mod/profiles/set/right/report_search.rb
```diff
@@ -80,16 +137,32 @@ def research_group_report_query
...
+      listing = render("#{cardtype}_sublist".to_sym)
+      return listing if card.research_group? || !subvariants
+      tab_listing listing
+    end
+
+    next if cardtype == :metric_value
+
+    view "#{cardtype}_sublist" do
+      card.variant = voo.structure if voo.structure
       default_listing
```
ethn: I like this but might suggest breaking it into more methods (eg `define_tab_listing_where_applicable`)

mod/profiles/set/self/research_group.rb
```diff
@@ -0,0 +1,19 @@
...
+  case variant
+  when :submitted
+    { created_by: user_id }
+  when :organized
+    { right_plus: [OrganizerID, { refer_to: user_id }] }
+  else
+    { or:
+        created_query(user_id, :submitted).merge(
+          created_query(user_id, :organized)
+        ) }
```
ethn, line 17: Have you checked performance on these?

mod/metrics/set/type_plus_right/metric_value/checked_by.rb
xithan: good point. I couldn't abandon the representation completely (because we need a "currently requested" state and a "was requested" state. I will explain that in the code). I have to update the migration.

mod/metrics/set/type/metric/creation.rb
xithan: no, that's just there so that it's more likely that new data is created correctly but I didn't want to deal with new issues (like failing imports because of existing non-metric title cards)

mod/profiles/set/self/research_group.rb
xithan: Doesn't seem very heavy to me that search. Tab switching on staging is pretty fast.

PR #622, opened 2017-03-10
badges

xithan:


db/migrate_cards/20170303130758_update_answer_table.rb
```diff
@@ -0,0 +1,8 @@
+# -*- encoding : utf-8 -*-
+
+class UpdateAnswerTable < Card::Migration
+  disable_ddl_transaction!
```
ethn, line 4: cool

db/migrate_cards/20170309143754_award_badges.rb
```diff
@@ -0,0 +1,155 @@
...
+    # #, :source, :metric, :wikirate_company
+    # ].each do |type_code|
+    #   award_create_badges type_code
+    # end
+    # binding.pry
+    # card = Card.fetch "some random card", new: {}
+    # card.save!
+    #card = Card.fetch "Philipp Kuehl+Project+Badges earned", new: { type_id: Card::PointerID }
+    #card.add_badge_card Card.fetch("Project Launcher")
+    #card.save!
```
ethn, line 19: is all this just commented for now?

mod/badges/set/abstract/02_badge_hierarchy.rb
```diff
@@ -0,0 +1,119 @@
...
+    lambda do |user_id|
+      user = user_id ? Card[user_id] : Auth.current
+      vote_card_ids = [user.upvotes_card.id, user.downvotes_card.id].compact
+      next nil unless vote_card_ids.present?
+      {
+        type_id: type_id,
+        referred_to_by: { id: ["in"] + vote_card_ids }
+      }
+    end
+  end
```
ethn, line 61: this isn't already available elsewhere?

mod/layout/set/abstract/thumbnail.rb
```diff
@@ -9,12 +9,12 @@
     end
   end
 
-  view :thumbnail do
+  view :thumbnail do |args|
```
ethn, line 12: what's your policy about "args" now?

ethn (approved):
Wow.  Is it a sign that our API is maturing that you're able to be this productive with it now, or have you just mastered the art of riding a wild bull?

db/migrate_cards/20170309143754_award_badges.rb
xithan: I moved it to a script because the integration stage is not executed in migrations. I added a PT story for that.

mod/layout/set/abstract/thumbnail.rb
xithan: The wild voo bull always throws me off when I need custom arguments so I use args when my parameter is not in the list of parameters that voo supports

mod/badges/set/abstract/02_badge_hierarchy.rb
xithan: probably possible to reuse the counts of the tabs on the profile page but not worth the effort since we are going to introduce a new caching system there

xithan:
That's just luck. Sooner or later the bull strikes back and drives you mad with weird bugs. Just happened again.

mod/layout/set/abstract/thumbnail.rb
xithan: so how can I pass arguments to another view with voo? I don't see that implemented

PR #625, opened 2017-03-21
improved company mapping

xithan:
using @vasgat's company-mapping gem


xithan:
hm, reopen doesn't update the diff. Looks like I have to close and resubmit it

PR #633, opened 2017-03-24
Ux improvements 6

srivig:
- fixed the missing browse page filter styles
- thumbnail_plain with link


xithan (approved):
You can ignore the failing source test. It fails because it uses a file on wikirate.org that is currently not available.

PR #639, opened 2017-03-29
badges fixes and cached counts

xithan:
didn't work properly on imports


mod/badges/set/abstract/award_badges.rb
```diff
@@ -20,10 +20,14 @@ def earns_badge action
...
   name_parts = [Auth.current, badge_card.badge_type, :badges_earned]
-  badge_pointer =
-    subcard(name_parts) ||
-      attach_subcard(Card.fetch(name_parts, new: { type_id: PointerID }))
+  badge_pointer = Card.fetch(name_parts, new: { type_id: PointerID })
+  if ActManager.include? badge_pointer.name
+    director = ActManager.fetch(badge_pointer)
+    director.reset_stage
+    badge_pointer = director.card
+  end
```
ethn, line 28: maybe separate into #badge_pointer method?

xithan:
I pushed accidentally cached count changes to this PR

mod/badges/set/abstract/award_badges.rb
ethn: why does fetch not find the correct card?  Isn't it supposed to?

I've been thinking that the ActManager should be pretty much invisible to mod developers in most cases.  Is that how you think about it?

mod/metric_answer_lookup/lib/answer.rb
```diff
@@ -74,22 +74,27 @@ def importance_sort mas, args
...
+      puts "failed: #{ma_id}"
+    end
+
+    def refresh_all fields
+      count = 0
+      Card.where(type_id: Card::MetricValueID).pluck_in_batches(:id) do |batch|
+        count += batch.size
+        puts "#{batch.first} - #{count}"
+        refresh(batch, *fields)
       end
```
ethn, line 109: much better!!

mod/metrics/spec/set/type/metric_value_import_file_spec.rb
```diff
@@ -77,8 +78,41 @@ def run_import
...
+      expect(badge_names).not_to include "Apple Inc.+Researcher+company badge"
+      Card::Env.params[:import_data] = []
+      add_three_answers "Apple Inc."
+      add_three_answers "Samsung"
+      run_import
+
+      expect(badge_names)
+        .to include "Apple Inc.+Researcher+company badge",
+                            "Samsung+Researcher+company badge"
+    end
```
ethn: +1

mod/wikirate/spec/set/type/pointer/pointer_spec.rb
```diff
@@ -1,134 +1,122 @@
...
-                               type_id: Card::PointerID,
-                               content: "[[#{@collection_card.name}]]"
-      big_blind.format(:json).render_export
+    let(:elbert) do
+      create "Elbert Hubbard",
+             content: "Do not take life too seriously."
+    end
+    let(:elbert_punchline) do
+      create "Elbert Hubbard+punchline",
+             content: "You will never get out of it alive."
```
ethn, line 11: nice examples!

mod/wikirate/spec/set/type/pointer/pointer_spec.rb
```diff
@@ -1,134 +1,122 @@
...
-                               content: "You will never get out of it alive.")
+        expect(json_export)
+          .to include(
+                { name: "normal pointer",
+                  type: "Pointer",
+                  content: "[[Elbert Hubbard]]\n[[Elbert Hubbard+punchline]]" },
+                { name: "Elbert Hubbard",
+                  type: "Basic",
+                  content: "Do not take life too seriously." },
+                { name: "Elbert Hubbard+punchline",
```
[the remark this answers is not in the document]
ethn: I disagree with the hound here (and agree with you)

ethn (commented):
I would be interested in hearing why the ActManager stuff is necessary.  It's a clever fix, but I'm wondering if we can find a way to solve this problem by improving ActManager or caching or something.

(I'm guessing you will persuade me that's not practical yet and we will merge as is, but I wanted to ask anyway)

mod/badges/set/abstract/award_badges.rb
xithan: I'm not sure if `badge_pointer = director.card` is really necessary. I think I added that before I found the real issue which is that on metric value imports we add all metric values in the integrate stage of the import card. When the first metric value adds a badge to the badge_pointer the badge_pointer catches up to the integrate stage and all badges added later to the badge_pointer by other metric values of that import are not saved (because the storage phase for the badge_pointer is already completed)

mod/wikirate/spec/set/type/pointer/pointer_spec.rb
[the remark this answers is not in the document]
xithan: Maybe it's time to say good bye to our hound. Hard to find the real comments between all those woofs. Codeclimate has improved a lot and does an at least equally good maybe even better job of marking style violations.

db/migrate_cards/20170330102819_cached_count_table.rb
```diff
@@ -0,0 +1,19 @@
+# -*- encoding : utf-8 -*-
+
+class CachedCountTable < ActiveRecord::Migration
+  def up
+    drop_table :counts if table_exists? :counts
+    create_table :counts do |t|
+      t.integer :left_id
+      t.integer :right_id
+      t.integer :value
+    end
```
ethn, line 10: so simple.  thing of beauty.

db/migrate_cards/20170401182741_remove_cached_count_cards.rb
```diff
@@ -0,0 +1,36 @@
...
+    remove type: :wikirate_company, plus_right: :analyses_with_articles
+    remove type: :wikirate_topic, plus_right: :source
+    remove type: :wikirate_topic, plus_right: :claim
+    remove type: :wikirate_topic, plus_right: :analyses_with_articles
+    remove type: :wikirate_topic, plus_right: :metric
+    remove type: :wikirate_topic, plus_right: :project
+    remove type: :wikirate_topic, plus_right: :wikirate_company
+  end
+
+  def remove type: , plus_right:
```
ethn, line 24: what do these colons mean?

mod/badges/set/abstract/answer_create_affinity_badge.rb
```diff
@@ -31,6 +31,14 @@ def virtual?
...
     end
   end
+
+  view :name_with_certificate do
+    "#{certificate(badge_level)} <strong>#{affinity}</strong> #{badge}"
+  end
+end
+
+def badge
+  cardname.parts[1]
```
ethn, line 41: why does this file both delegate `badge` to card _and_ define badge?

mod/badges/set/abstract/award_badges.rb
ethn: So do you think we can take out the ActManager part?

If it stays in, I think we should be able to explain the cases in which a mod developer would be expected to use this kind of thing, no?

mod/caching/spec/set/type_plus_right/metric/source_spec.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "../../../support/cached_count_shared_examples"
```
ethn, line 1: I like these shared examples!!

mod/wikirate/spec/set/right/wikirate_link_spec.rb
```diff
@@ -8,7 +8,7 @@
   # end
   # it "should block changing the url for existing source" do
   #   # create the page with source
-  #   url = "http://www.google.com/?q=wikirateissocoolandawesomeyouknowsigh"
+  #   url = "http://www.google.com/?q=wikiratesigh"
```
ethn, line 11: haha.  I think this may have slightly changed the meaning.

ethn (approved):
great stuff.  Still a comment or two worth attention, but ready to roll

db/migrate_cards/20170401182741_remove_cached_count_cards.rb
xithan: required keyword arguments (https://robots.thoughtbot.com/ruby-2-keyword-arguments)

mod/badges/set/abstract/award_badges.rb
xithan: we can try to let `attach_subcard` handle that issue

mod/wikirate/spec/set/right/wikirate_link_spec.rb
xithan: that was a global search and replace of `wikirateissocoolandawesomeyouknow` :smile:

mod/badges/set/abstract/answer_create_affinity_badge.rb
xithan: it delegates `badge` from format to card

PR #640, opened 2017-03-30
Various bux fixes and style improvements

srivig:
- Fixed formula view style bugs
- Better 'source preview' page 
- 'Source preview' code cleanup
- Source page default view to `:preview` and source details structure to haml template view
- Updated spec
- Fixed collapse js
- Feature to remove a metric when editing a WikiRating formula
- Added "About me" and "Discussion" section on profile page


mod/metrics/set/type_plus_right/metric/formula.rb
```diff
@@ -78,11 +78,11 @@ def add_metric_button
   view :core do
     return _render_rating_core if card.wiki_rating?
     return _render_categorical_core if card.categorical?
-    "= #{super()}"
+    "<span>=</span><span>#{super()}"
```
xithan: the closing `span` is missing

xithan (changes requested):

ethn:
looks good (with the small change xithan requested)

mod/metrics/set/type_plus_right/metric/formula.rb
srivig: fixed it and it will be available in my next push (along with source preview page related updated )

mod/wikirate_source/set/type/source/source_preview.rb
```diff
@@ -159,131 +143,15 @@ def valid_mime_type? mime_type
...
         <a href="#{file_card.attachment.url}" class="btn btn-primary" role="button">Download</a>
       HTML
     else
-      url_card = card.fetch(trait: :wikirate_link)
-      url = url_card ? url_card.item_names.first : nil
       <<-HTML
         <a href="#{preview_url}" class="btn btn-primary" role="button">Visit Original Source</a>
       HTML
     end
   end
```
xithan: since you refactored/removed all that other stuff no reason to stop here and save that ugly piece
For example:
```
file_card = card.fetch trait: :file
url, text = file_card ? [file_card.attachment.url, "Download"] : [preview_url, "Visit Original Source"]
link_to text, href: url, class: "btn btn-primary", role: "button"
```

xithan (approved):

xithan:
If you merge with the latest master semaphore might pass.

db/migrate_cards/data/cards/source_structure
```diff
@@ -0,0 +1,32 @@
...
+		<div class="col-md-11 column">
+		    {{+Companies|titled; title:Companies|link}}
+        {{+Topics|titled; title:Topics|link}}
+        {{#+Tags|titled;title:Tags; item:link}}
+		</div>
+	</div>
+	<div class="row clearfix">
+	  {{+Discussion|titled;title:Discussion;show:comment_box,menu}}
+	</div>
+</div>
```
xithan: hound can't recognize that but html structures in cards is considered to be bad style. I suggest you make it a view and convert it to haml (e.g with http://htmltohaml.com) and replace nests like ` {{+Year|titled;title:Year|name}}` with `= field_nest :nest, view: :titled, title: "Year", items: { view: :name}`

mod/wikirate_source/set/type/source/tabs.rb
```diff
@@ -0,0 +1,42 @@
...
+    {
+      details_tab: "#{fa_icon :info} Details",
+      metrics_tab: "#{fa_icon 'bar-chart'} #{metric_count} Metrics",
+      notes_tab: "#{fa_icon 'quote-left'} Notes",
+      download_tab: { html: download_tab_link }
+    }
+  end
+  view :details_tab do
+    tab_wrap do
+      render_core structure: "source_structure", show: "header"
```
xithan: Does `show: "header"` do something here? 
I suggest ` tab_wrap { _render_details }`,  add view `view :details, template: :haml` and move the structure to `wikirate_source/template/set/type/source/tabs/details.haml`

xithan (changes requested):

mod/wikirate_source/set/type/source/source_preview.rb
```diff
@@ -148,142 +132,22 @@ def valid_mime_type? mime_type
...
-      case card.source_type_codename
-      when :wikirate_link
-        link_button url
-      when :file
-        file_download_button
-      else
-        ""
-      end
-    result
+    url, text = if file_card [file_card.attachment.url, "Download"]
```
xithan: Is this valid ruby? Condition and statement in the same line without any separation like line break or `then`

srivig: `then` is missing, I am wondering how my ide linter never pointed that

mod/wikirate_source/template/type/source/tabs/source_details.haml
```diff
@@ -0,0 +1,10 @@
+.col-md-12.column
+  .row.clearfix
+    = nest(card, view: :listing)
+  .row.clearfix
+    = field_nest :description, view: :titled, title: "Description"
+    = field_nest :year, view: :titled, title: "Year", items: { view: :name}
+    = field_nest "Report type", view: :titled, title: "Report Type", items: { view: :name}
+    = field_nest :wikirate_company, view: :titled, title: "Companies", items: { view: :link}
+    = field_nest :wikirate_topic, view: :titled, title: "Topics", items: { view: :link}
+    = field_nest :discussion, view: :titled, title: "Discussion", show: :comment_box
```
xithan: that's so much easier to read 👍

mod/wikirate_source/template/type/source/tabs/source_details.haml
```diff
@@ -0,0 +1,10 @@
+.col-md-12.column
+  .row.clearfix
+    = nest(card, view: :listing)
```
xithan: for nests of the same card you can call render:
`= _render :listing` (underscore for skipping permission checks)

mod/wikirate_source/template/type/source/tabs/source_details.haml
```diff
@@ -0,0 +1,10 @@
...
+  .row.clearfix
+    = field_nest :description, view: :titled, title: "Description"
+    = field_nest :year, view: :titled, title: "Year", items: { view: :name}
+    = field_nest "Report type", view: :titled, title: "Report Type", items: { view: :name}
```
xithan: better practice: add codename if you use a card name in code:
> wagn g card:migration add_codename_for_report_type
```
# db/migrate_cards/add_codename_for_report_type
ensure_card "Report type", codename: "report_type"
```

xithan (changes requested):
two little tweaks then it's perfect 🥇

xithan (approved):

mod/wikirate/lib/javascript/script_collapse.js.coffee
```diff
@@ -88,16 +88,16 @@ findCollapseTarget = (toggle) ->
...
 
-  target = $toggle.data('target') || '.collapse'
+  if $toggle.attr("href") != null
+    return $($toggle.attr("href"))
+  target = $toggle.attr("href") || $toggle.data('target') || '.collapse'
   if $toggle.data("collapse")?
     target += $toggle.data("collapse")
-
-  if $toggle.find(target).length
+  else if $toggle.find(target).length
```
xithan, line 96: If you change js or css then you have to update the test database (with `bundle exec  rake wikirate:test:seed:update`) otherwise the tests won't check your changes. The changes above broke the filter form toggle on metric and company pages.
Several points are suspicious there:
1) Line 95 changes `target` which is used in the next if block but you connected that if block with an `else` so that `target` is not used anymore. 
2) This method is supposed to return the element that we want to collapse. "href" seems to be an odd name  choice to contain that information. 
3) you return in line 92 if there is an href attribute but then you check agin in line 93 again for an href attribute

srivig: you are right, line 93 and line 96 are wrong. I think I forgot to delete/redo it. My intention is only to add these two lines only. 
`if $toggle.attr("href") != null`
 `   return $($toggle.attr("href"))`
Above snippet enables `data-collapse-text-in/out` to work for the bootstrap collapse button element where the target value is declared in `href` instead of `data-target`. In my opinion, declaring _target_ value on `href` is right way instead of in `data-target` attribute.

xithan: I checked the bootstrap docs. They use `href` for links and `data-target` for buttons so we should support both.

xithan: `$toggle.attr("href") != null` is always true. If it is not defined it returns `undefined` not `null`. It seems to work with `target = $toggle.attr("href") || $toggle.data('target') || '.collapse'`

srivig: `target = $toggle.attr("href") || $toggle.data('target') || '.collapse'` isn't working on browse metric page

srivig: maybe we could do this
 `href_attr = $toggle.attr("href")`
 ` if typeof href_attr != typeof undefined && href_attr != false`
 `  return $($toggle.attr("href"))`

PR #680, opened 2017-05-31
timeline clean up

xithan:
https://www.pivotaltracker.com/story/show/136576609


mod/metrics/set/type/metric_value/views.rb
```diff
@@ -143,17 +143,6 @@ def beautify value
...
-  view :timeline_data do
-    wrap_with :div, class: "timeline-row" do
-      [
-        _render_year,
-        _render_value,
-        _render_chart
-      ]
-    end
-  end
-
```
ethn, line 156: was this not in use?

ethn (approved):

mod/metrics/set/type/metric_value/views.rb
wagn: not anymore

PR #681, opened 2017-05-31
back to green

xithan:


mod/wikirate/features/company.feature
```diff
@@ -45,7 +45,7 @@ Feature: company feature
     And I select "Not Researched" from "Value"
     And I submit form
     Then I should not see "disturbances in the Force"
-    And I should see "Sith Lord in Charge"
+    And I should see "Weapons"
```
ethn, line 48: why did this change?

ethn (approved):
curious about the one question, but all looks good.

mod/wikirate/features/company.feature
wagn: because that metric has now a researched value. I added it because I needed a free text metric with an answer for another test.

PR #684, opened 2017-06-08
csv and json links 

xithan:
https://www.pivotaltracker.com/story/show/139559677


mod/wikirate_source/set/type/source.rb
```diff
@@ -7,6 +7,11 @@
 add_attributes :import
 attr_accessor :import
 
+def source_title_card
+  # FIXME: needs codename, but :title is linked to *title
```
ethn, line 11: could change *title to `*site title*` and change codename accordingly...

ethn (approved):
great stuff.

PR #690, opened 2017-06-20
vega update

xithan:
adapt to the api changes of the new vega version


db/migrate_cards/20170616220037_remove_d3_library.rb
```diff
@@ -0,0 +1,10 @@
+# -*- encoding : utf-8 -*-
+
+class RemoveD3Library < Card::Migration
+  def up
+    Card["script: libraries"].drop_item! "script: d3"
+    Card["script: d3"].update_attributes! codename: nil
+    Card::Cache.reset_all
+    Card["script: d3"].delete
```
xithan, line 8: d3 is now included in vega

ethn (approved):
Is it safe to assume that these changes are based on moving to the vega API and do not affect the graphs' appearance?

xithan:
no visible change (except that there is a chance that the bug with the magnifying glass cursor is fixed. I didn't check on that)

PR #696, opened 2017-06-27
wagn tick of OC widget

xithan:


mod/wikirate/lib/open_corporates/api.rb
```diff
@@ -20,7 +20,7 @@ def fetch_companies jurisdiction_code, company_number, opts={}
       # This method removes the additional hash level with the "jurisdiction" key.
       # @return [Array<Hash>]
       def fetch_jurisdictions
-        resupt =
+        result =
```
ethn, line 23: :)

mod/wikirate/lib/open_corporates/api.rb
```diff
@@ -34,11 +34,6 @@ def fetch_jurisdictions
       # @return the full json response converted to a hash
       def fetch *query_args
         JSON.parse json_response(*query_args)
```
ethn, line 36: what happens to parse errors?  (Looks like it may have been incorrectly rescuing those before...)

mod/wikirate/set/right/open_corporates.rb
```diff
@@ -63,6 +63,6 @@ def jurisdiction
   def incorporation_date
     date = oc.incorporation_date
     return "" unless date
-    "#{date.strftime '%d %B %Y'} (#{time_ago_in_words(date)} ago)"
+    "#{date.strftime '%-d %B %Y'} (#{time_ago_in_words(date)} ago)"
```
ethn, line 66: ?

ethn: (what's the dash?)

ethn (approved):
comments are from curiosity, not concern.

mod/wikirate/set/right/open_corporates.rb
xithan: https://apidock.com/ruby/DateTime/strftime - not zero-padded

mod/wikirate/lib/open_corporates/api.rb
xithan: Unlikely that if we get a response that it won't be json. Airbrake will tell us if I'm wrong

PR #709, opened 2017-11-29
import and formula fixes

xithan:


mod/metrics/set/right/variables.rb
```diff
@@ -74,13 +91,14 @@ def variable_row item_name, index, args
...
+  view :missing do
+    return super() unless card.new_card?
+    if card.formula_card
       card.extract_metrics_from_formula
-      render!(@denied_view, args)
+      render! @denied_view
     else
-      super(args)
+      Auth.as_bot { card.save! }
+      render! :core
```
ethn: probably should be `@denied_view`, not `:core`, no?

mod/metrics/set/right/variables.rb
```diff
@@ -48,7 +46,26 @@ def input_metric_name_by_index index
...
+        button_tag class: "pointer-item-add slotter",
+                   situation: "outline-secondary",
+                   data: { toggle: "modal", target: target },
+                   href: path(layout: "modal", view: :edit,
+                              mark: card.name,
+                              slot: { title: "Choose Metric" }) do
+          fa_icon(:plus) + " add metric"
+        end
+    end
+  end
```
ethn: I like this move.

lib/capistrano/tasks/delayed_job.cap
```diff
@@ -0,0 +1,25 @@
+namespace :delayed_job do
+  desc "stop delayed job processes"
+  task :stop do
+    on roles(:app) do
+      within release_path do
+        execute "pwd"
+        env = "env RAILS_ENV=production"
+        num_jobs = fetch :num_delayed_job_processes
+        execute :rvm, "default do", env, "./bin/delayed_job -n", num_jobs, "stop"
```
ethn, line 9: I don't love how I did this, but it wasn't working if I called it without the rvm nonsense.

ethn: if methods work in .cap, this would be a nice use case!

mod/metrics/set/right/variables.rb
```diff
@@ -36,10 +36,8 @@ def input_metric_name_by_index index
   item_cards.fetch(index, nil).name
 end
 
-
-
 format :html do
-  view :core do |args|
+  view :core, chache: :never do |args|
```
ethn: typo

ethn: why wouldn't caching work here?

ethn (changes requested):
some small changes requested.

This has all the code that you added to production today, yes?

mod/metrics/set/right/variables.rb
xithan: I wanted to make sure that that wasn't the reason that updates on this card had no effect

ethn: I'm going to guess `chache: :never` didn't fix it :)

ethn: FYI, view caching is off in development mode by default, so if it's breaking in development (and you haven't explicitly turned it on) it's probably not view caching's fault

xithan:
I'll try again to get semaphore to work

PR #718, opened 2017-12-01
Fix relationship metrics

xithan:
#152277747


mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
```diff
@@ -17,7 +17,7 @@ def company_name
   Answer.select(:company_id).where(metric_id: metric_id).uniq
         .pluck(:company_id).map do |company_id|
     # faster way to get this from company+topic?
-    Card.fetch(company_id.cardname.trait :wikirate_topic)
+    Card.fetch(company_id.cardname.trait(:wikirate_topic))
```
ethn, line 20: do we have a style preference for these situations?

mod/csv_import/set/abstract/import/01_table_row.rb
```diff
@@ -104,7 +103,7 @@ def respond_to_missing? method, _include_private=false
   end
 
   def field_from_method method
-    method =~ /^(\w+)_field$/ ? $1 : nil
+    method =~ /^(\w+)_field$/ ? Regexp.last_match(1) : nil
```
ethn, line 106: I'm not in love with last_match, but I can't figure out how to do this common pattern much more cleanly.

```
m = method.match /^(\w+)_field$/
m && m[1]
```
I wish this worked: `method.match(/^(\w+)_field$/)&[1]

ethn (approved):
Whenever there are this many changes, any help getting to the meat of the main logic changes would help.  I skimmed the whole thing, and the relationship changes look good from what I can see.

mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
xithan: inner parenthesis are mandatory, outer ones are optional

mod/csv_import/set/abstract/import/01_table_row.rb
xithan: how about `method.match(/^(\w+)_field$/)&.captures&.at(0)`?

xithan:
You can use the commit links to get rid of all the clean up changes.

mod/csv_import/set/abstract/import/01_table_row.rb
ethn: it's actually more readable than my "wish" but probably less so than either the current code or my attempt. 

We could also add our own method to String:
```
def nth_match n, regexp
...

method.nth_match 1, /^(\w+)_field$/
```

xithan: I don't like that you first say in what part of the match you are interested in and then you define and do the match. 
We could add `def capture n` to MatchData instead:
`method.match(/^(\w+)_field$/)&.capture(0)`

xithan: https://github.com/decko-commons/decko/pull/129/files

PR #720, opened 2017-12-04
fix import tool

xithan:
- fix issue with duplicate detection
- company alias matching wasn't working work
- new start page for import tool with tabs for different import types


mod/csv_import/template/self/import_tool.haml
```diff
@@ -0,0 +1 @@
+= tabpanel
```
ethn, line 3: is this worth a separate file?

ethn (approved):

mod/csv_import/template/self/import_tool.haml
xithan: Not anymore, but is it worth to change?

ethn: nah

PR #721, opened 2017-12-05
project page colours

srivig:
• colour for left column links for companies and metrics.
• colour for status number links within progress bars for companies/metrics/years
• colour for status bar in years (unknown is same as known)


mod/projects/set/ltype_rtype/metric/project.rb
```diff
@@ -44,6 +44,6 @@ def project_name
 
   def metric_link values=:all
     path_args = card.project_card.filter_path_args values
-    link_to_card card.metric_card, yield, path: path_args
+    link_to_card card.metric_card, yield, path: path_args, class: "metric-color"
```
ethn, line 47: seems like an iffy class name. maybe `metric-filtered-link`

mod/projects/set/type_plus_right/project/year.rb
```diff
@@ -37,6 +37,6 @@ def year_progress_table
                    card.all_year_project_cards,
                    [:fancy_year, :research_progress_bar],
                    header: ["Year", "Answers Researched"],
-                   td: { classes: ["year-answer"] }
+                   td: { classes: ["year-answer", "overall-progress-box"] }
```
ethn: does this mean the year will have the same color as the overall progress bar?

mod/wikirate/lib/stylesheets/style_wikirate_progress_bar.scss
```diff
@@ -14,6 +14,16 @@
   font-weight: bolder;
 }
 
+.progress {
+  .progress-known, .progress-unknown {
+    a {
+       color: $white-text-color;
+    }
+  }
```
ethn, line 22: should there be something here that separates known and unknown?

ethn (changes requested):
I would merge this but semaphore tests are failing.  Did you intend to update the seed db?

mod/projects/set/ltype_rtype/metric/project.rb
srivig: `metric-color` is a generic class name, I prefer to re-use CSS classes instead of creating a new one.

mod/projects/set/type_plus_right/project/year.rb
srivig: `overall-progress-box` has default progress bar colours. Maybe I can  rename the class name with more generic term

mod/wikirate/lib/stylesheets/style_wikirate_progress_bar.scss
srivig: It is just white text on dark background and dark text in the light background. I don't think if we need to differentiate link colours more specifically.

mod/projects/set/ltype_rtype/metric/project.rb
ethn: ok

mod/projects/set/type_plus_right/project/year.rb
ethn: cool

mod/wikirate/lib/stylesheets/style_wikirate_progress_bar.scss
ethn: this has white for both known and unknown.  One big problem was that on staging known and unknown had the same background color.  I was asking whether there should be a general purpose fix.

ethn (approved):

PR #722, opened 2017-12-06
bug fixes for calculated metrics

xithan:


mod/metrics/set/abstract/metric_thumbnail.rb
```diff
@@ -14,23 +15,24 @@ def thumbnail_image
   end
 
   def thumbnail_title
-    content = wrap_with(:div, nest(card.metric_title_card,
-                                   view: :name),
-                        class: "ellipsis")
-    return content unless voo.show?(:thumbnail_link)
+    return thumbnail_title_text unless voo.show?(:thumbnail_link)
```
ethn: fwiw, I'm increasingly favoring the pattern where the top level method has something like:
```
voo.show?(:thumbnail_link) ? thumbnail_title_link : thumbnail_title_text
```
a lot of times some of the most important logic is hidden at the end of a return statement :)

ethn: not necessarily asking for that change here and now.

ethn (approved):
This looks good.  It looks like there are a couple of codeclimate issues.  If you address those, can you also do a decko tick?  (when we can avoid separate PRs for those, it makes things faster!)

PR #725, opened 2017-12-08
value type tweaks

xithan:
get rid of the button in the metric details table and show the value type like the other properties.


mod/metrics/set/abstract/researched.rb
```diff
@@ -1,115 +0,0 @@
-def score_cards
```
xithan, line 1: this was a duplicate of `metrics/set/abstract/01_researched.rb` with a few fixes.
`abstract/relationship` included `01_researched.rb` without the fixes and then `researched.rb`  overrode it. I moved all the changes to `01_researched.rb`

mod/metrics/set/abstract/04_chart.rb
```diff
@@ -63,7 +63,7 @@ def vega_chart
     output [
       zoom_out_link,
       wrap_with(:div, "",
-                id: id, class: classy("vis"),
+                id: id, class: "#{classy("vis")} _load-vis",
```
xithan: When we open a modal on a metric page the slot of the main pages receives a "slotReady" for every slot in the modal and loads every time the visualization. With this `_load-vis` class that I remove when it is loaded we save at least at l lot of unnecessary request.

mod/metrics/set/abstract/01_researched.rb
```diff
@@ -25,55 +25,68 @@ def tab_list
       wrap_with :div, class: "metric-details-content" do
```
xithan, line 25: All changes here are old but they were in the wrong file.

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -41,28 +41,6 @@ $(window).ready ->
...
-        'number_details'
-      when 'Money'
-        'currency_details'
-      when 'Category', 'Multi-Category'
-        'category_details'
-      else
-        ''
-  hideAll(slot)
-  showField(div_to_show)
-
```
xithan, line 65: logic for value type moved to separate file

mod/metrics/set/type_plus_right/metric/value_type.rb
```diff
@@ -59,12 +59,16 @@ def default_edit_args _args
...
   end
 
-  def single_card_edit_slot
-    multi_card_edit_slot
+  def single_card_edit_field
+    super() + fields_form
+  end
+
+  def editor_in_multi_card
+    super() + fields_form
```
xithan, line 71: value type editor was working for the last four months because these methods were renamed in decko.

xithan: very hacky, we need a clearer way to add additional stuff to all kind of editors. Probably a edit structure, but using self in the edit structure needs other hacky stuff to prevent loops.

ethn (approved):

PR #726, opened 2017-12-11
calculated metric fixes

xithan:


mod/csv_import/spec/support/shared_csv_import.rb
```diff
@@ -107,28 +132,28 @@ def with_row index:, context:, checked:, fields:, match:, suggestion:
 # define :company_row and :value_row to use
 # the helper to access the original import data
 shared_context "answer import" do
-  def answer_card key
-    Card[metric, company_name(key), year]
+  def answer_name key, override={}
+    [metric, company_name(key, override), year].join "+"
```
ethn: Card::Name[metric, company_name(key, override), year]

mod/layout/set/abstract/media.rb
```diff
@@ -4,7 +4,8 @@ def image_card
   end
 
   def image_src opts
-    image_card&.format&.render_source(size: opts[:size])
+    return "" unless image_card
+    nest image_card, view: :source, size: opts[:size]
```
ethn: There is a potential problem I was just dealing with elsewhere (and maybe related to your question yesterday).

If I do this:
```
attribute_value = nest cardname, viewname
wrap_with :div, attribute: attribute_value...
```
...then there is the potential for the nest to return a stub and the stub to get escaped as it's put inside the attribute.

possibilities:
1. find a way to avoid the escaping (haml, maybe?)
2. create the view on the image
3. render the image source without a subformat (eg image_card.format.render...).  This will work, but the cache won't get cleared if the image is changed

ethn: in other words, the old code may be safer :(

mod/projects/set/type_plus_right/project/metric.rb
```diff
@@ -27,7 +27,7 @@ def default_item_view
   end
 
   def editor
-    :filtered_list
```
ethn: no longer needed

mod/projects/set/type_plus_right/project/wikirate_company.rb
```diff
@@ -27,7 +27,7 @@ def default_item_view
   end
 
   def editor
-    :filtered_list
+    :list
```
ethn: no longer needed (assuming we merge my stuff)

test/shared_data/calculated_metrics.rb
```diff
@@ -0,0 +1,53 @@
...
+    def wikirating_metrics
+      Card::Metric.create(
+        name: "Jedi+darkness rating",
+        type: :wiki_rating,
+        formula: { "Jedi+deadliness+Joe User" => 60,
+                   "Jedi+disturbances in the Force+Joe User" => 40 }
+      )
+    end
+  end
+end
```
ethn, line 53: nice

ethn (approved):
wow, good stuff.

PR #727, opened 2017-12-11
fix styling of filtered list items

ethn:
much simpler one :)


mod/csv_import/set/right/import_status.rb
```diff
@@ -46,8 +46,8 @@ def save_status
   delegate :status, :import_counts, to: :card
   delegate :percentage, :count, :step, to: :import_counts
 
-  def wrap_data
-    super.merge "refresh-url" => path(view: @slot_view)
+  def wrap_data slot=true
+    super(slot).merge "refresh-url" => path(view: @slot_view)
```
xithan: you need the argument for `super` only in blocks not in methods

xithan (approved):

xithan (approved):
👍

PR #729, opened 2017-12-12
more thumbnail fixes

xithan:


mod/csv_import/set/abstract/import/table_row_relationship.rb
```diff
@@ -10,8 +10,8 @@ class TableRowRelationship < TableRowWithCompanyMapping
   def initialize csv_row, format
     super
     @related_match = match_company(:related_company)
-    @related_wikirate_company, @related_match_type = @related_company_match.match
-    @related_company = @related_company_match.suggestion
+    @related_wikirate_company, @related_match_type = @related_match.match
```
ethn: some comments might be helpful here.  not super easy to guess what `x, y = z_match.match` is all about.

mod/layout/set/abstract/thumbnail.rb
```diff
@@ -1,20 +1,20 @@
 format :html do
-  view :thumbnail_plain do |args|
+  view :thumbnail_plain, cache: :never do |args|
```
ethn: none of these should be necessary. 

We can't just keep slapping `cache: never` everywhere.  If there's a bug in the view caching we have to fix it.

ethn (changes requested):
you can assign this to me if you want, but I'm not ok with using `cache: :never` when there's no good reason view caching shouldn't work.

mod/layout/set/abstract/thumbnail.rb
xithan: My understanding was that we can't use view caching if we use args.

xithan:
We can remove the `cache: :never` when you have a better solution but for now I'd like to get this out to fix the broken views on the live site.

mod/layout/set/abstract/thumbnail.rb
ethn: ah, is this all because of foreign args in the thumbnail_subtitle?

Couldn't we make it much cleaner and (by keeping caching) faster if we override the subtitle view in the metric type

ethn: In other words, I can believe that caching is not working as currently implemented, but I don't understand why we would implement this in a way that makes caching not work. This is one of the most used views on the site, and it seem like it would be worth getting caching working with it.

ethn (approved):

mod/layout/set/abstract/thumbnail.rb
```diff
@@ -1,28 +1,28 @@
...
 
-  view :thumbnail do |args|
+  view :thumbnail do
     voo.show :thumbnail_link
-    wrap_with :div, thumbnail_content(args), class: "thumbnail"
+    wrap_with :div, thumbnail_content, class: "thumbnail"
   end
 
-  view :thumbnail_no_link do |args|
+  view :thumbnail_no_link do
```
ethn, line 17: sooo much prettier!

ethn (approved):

PR #734, opened 2017-12-18
Subprojects

ethn:


mod/wikirate/lib/javascript/script_wikirate_coffee.js.coffee
```diff
@@ -124,6 +124,23 @@ $(window).ready ->
...
+decko.slotReady (slot) ->
+  return unless slot.hasClass("TYPE-project") && slot.find("form")
+  parent = slot.find(".RIGHT-parent .pointer-item-text")
+  appendParentToAddItem parent
+
+appendParentToAddItem = (parent) ->
+  return unless parent.val()
+  anchor = parent.slot().find "._add-item-link"
+  new_href =  anchor.attr("href") + "&" + $.param({ "filter[project]" : parent.val() })
+  anchor.attr "href", new_href
```
ethn, line 142: Not sure I like this solution.  We're going to need filter presets in many of our use cases, and this is a vary narrow way to achieve that.

mod/wikirate/set/all/helper.rb
```diff
@@ -3,7 +3,7 @@ def with_header header, level: 3
     output [wrap_with("h#{level}", header), yield]
   end
 
-  def standard_pointer_nest codename
+  def standard_nest codename
```
ethn, line 6: was already being used with non-pointers...

xithan (approved):

PR #738, opened 2017-12-21
Calcs

ethn:


mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -7,39 +7,19 @@ decko.editorContentFunctionMap['.pairs-editor'] = ->
   JSON.stringify(hash)
 
 $(window).ready ->
-  $('body').on 'input', 'input.metric-weight', (event) ->
-    result = 0
-    $(this).closest('tbody').find('input.metric-weight').each ->
```
ethn, line 12: turns out this was broken.  (there is no `input.metric-weight`)

mod/metrics/set/right/variables.rb
```diff
@@ -14,21 +14,13 @@ def formula_card
...
-def extract_metrics_from_formula
-  metrics = formula_card.input_names
-  Auth.as_bot do
-    update_attributes! content: metrics.to_pointer_content,
-                       type_id: PointerID
-  end
-  metrics
+def content
+  @content ||=
+    db_content.present? ? db_content : formula_card.input_names.to_pointer_content
```
xithan, line 19: This is a session card. `db_content` is always empty. The content is stored in `Env.session[key]`

xithan (changes requested):

mod/filter/lib/card/metric_filter_query.rb
```diff
@@ -1,6 +1,8 @@
 class Card
   class MetricFilterQuery < Card::FilterQuery
     def wikirate_topic_wql topic
+      topic = Array.wrap topic
+      topic.unshift :in if topic.length > 1
```
xithan: ideally wql would handle this

mod/metrics/set/right/variables.rb
ethn: well, I agree the nomenclature isn't clear, but it's not true that db_content is always empty, sometimes it gets set by `card[content]` param (eg, right after someone clicks "add selected" in the new editor).  I will make a note of this in the comments.

Now that you mention it, though, I'm actually not clear on whether it should still be a session card.  In the way I'm using `+*variables` with the new editor, it doesn't really need to be stored at all.  And I'm a little confused about why, if these were session cards, there are 7 `+*variable` cards in the (live) database: http://wikirate.org/*variables+*right+*count

xithan: I think the session card makes sense to keep the variable names consistent during a session. Otherwise the numbers can change if you save and edit again. 
The variables card accessor is a session card by default. Looks like sometimes they are accessed on a different way.

ethn: ok good points. So is there anything in the current functionality I have to change to move forward? I think most of the things that would have committed this to the database have been removed, so even if it's not accessed via the accessor, it shouldn't get saved to the db.

xithan: I don't know how it works know with the new interface but either we don't make any saves at all which means inconsistent variables name or we still save changes but that can't work because you override here the method that fetches the content from the session.

ethn: OK, that makes sense.

But I'm struggling to reconcile this with the need to delete variables from the list: https://www.pivotaltracker.com/story/show/153401058.

So if it's a session card and we delete a variable from the list, we have to have some special representation that marks an index number as unused. If it's not a session card, we have inconsistent variable names.  My question: is this really a problem?

Let's say I have M1, M2, M3, and then I use M1 and M3 in the formula and save.  Once it's saved, M1 and M3 won't show up with those names in the formula, they'll have their real names.  This much is true whether or not +*variables is a Session card.

If it's not a session card, then the metric not used in the formula disappears from the list, which is already almost as good as deleting.  In this version, we could also have a "delete" button that just hides the row.  That way the variable numbers wouldn't change.

If it's a session card, I'm just not sure how deleting works.  The hiding thing only works across multiple edits if we have a way to represent the row as hidden in the session card (which would require more than the pointer representation).

Short version: I agree that inconsistent variable names is not ideal, but I'm not actually convinced it's as big of a problem as the other ones that Session cards create.

xithan: A few more thoughts
1) The delete button is not really necessary. It's just a list of shortcuts that you create. It doesn't really matter if you have a shortcut there that you don't need anymore. You can just leave it.
2) There is more inconvenience if we don't use session cards. The order changes also.
If you save a formula `{{M2}}/{{M1}}` and then realize the quotient was the wrong way then you have to save it again as `{{M2}}/{{M1}}` to change the order because the order of the variables has changed to.
3) The crucial question is if there will be a lot of experimenting with a formula. For example if you try with different CO2 emission metrics and scores and check results before you decide which metrics to use it could be annoying that your selection of metrics disappears after every try. But maybe the new filter interface is so good that it's not such a big problem.
4) Not sure if this little feature is worth a long discussion. I'll approve this now and leave the decision to you

xithan (approved):

mod/metrics/set/right/variables.rb
ethn: A couple more points on this arguably-already-too-long conversation

re #1, I see your point, but Alex and others have expressed confusion about not being able to delete items.  Maybe help text could solve that?

re #2, after you save a formula, you probably won't use M1 or M2 again (unless you need to use the same variable more than once).  The nest will show up with the full name of the metric, and most will probably just cut and paste those around.

As discussed, I'm going to deploy this to staging and then discuss further with alex as needed.

PR #739, opened 2017-12-22
virtual calculated values

xithan:


mod/metric_answer_lookup/lib/answer/entry_fetch.rb
```diff
@@ -14,7 +14,7 @@ def fetch_metric_id
     end
 
     def fetch_record_id
-      card.left_id
+      card.left_id || card.left.id
```
ethn, line 17: realizing that perhaps left_id could be smart enough to look itself up...

ethn: (some day)

mod/metric_answer_lookup/lib/answer.rb
```diff
@@ -8,19 +8,35 @@ class Answer < ActiveRecord::Base
...
+  end
+
+  def virtual_answer_card name=nil, val=nil
+    name ||= [record_name, year.to_s]
+    val ||= value
+    Card.new(name: name, type_id: Card::MetricValueID).tap do |card|
+      card.define_singleton_method(:value) { val }
+      card.define_singleton_method(:value_card) do
+        Card.new name: [name, :value], content: val
+      end
```
ethn: we might consider adding all these virtual card methods to a module and having the singleton_class include it. My hunch is that it would be faster (performance-wise)

mod/metric_answer_lookup/lib/answer.rb
```diff
@@ -69,6 +113,20 @@ def respond_to_missing? *args
...
+    val.to_d
+  end
+
+  def unknown? val
+    val.casecmp("unknown").zero?
+  end
+
+  # true if there is no card for this answer
+  def virtual?
+    card&.new_card?
```
ethn: so nil if card is nil?

mod/metrics/set/abstract/value.rb
```diff
@@ -24,6 +24,10 @@ def company_key
   company.to_name.key
 end
 
+def company_id
```
ethn, line 27: maybe this should be called company_mark?

mod/metrics/lib/formula/calculator/value_store.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      def get card_id, company, year=nil
+        dig_args = [card_id, (company if @type[card_id] == :metric_value), year].compact
+        values(card_id).dig(*dig_args)
+      end
+
+      def add card_id, *args
+        value = args.pop
+        year = args.pop
+        values(card_id).dig(card_id, *args).merge!(year.to_i => value)
```
ethn, line 24: cool

ethn (approved):
good stuff.

there's a lot of good code in calculations.  glad it's going to see more use soon.

mod/metrics/set/abstract/value.rb
xithan: The code is wrong, not the name. Looks like I hit "undo" one time too often when I removed what I did after I corrected this.

mod/badges/spec/support/award_answer_create_badges_shared_examples.rb
```diff
@@ -15,21 +15,21 @@
   end
 
   describe "badges earned pointer" do
-    it "exists" do
-      expect(badge_pointer).to be_instance_of Card
-    end
+    # it "exists" do
+    #   expect(badge_pointer).to be_instance_of Card
+    # end
```
ethn: what changed here?

mod/caching/set/abstract/answer_table_cached_count.rb
```diff
@@ -51,7 +51,7 @@ def return_arg val
 
 # needed for "found_by" wql searches that refer to search results
 # of these cards
-def wql_hash
+def wql_from_content
```
ethn, line 54: there are lots of other places still overriding wql_hash rather than wql_content.  Do those not work?  I would think the caching wouldn't be an issue if you override wql_hash, no?

mod/caching/set/all/metric_answer_cache.rb
```diff
@@ -1,26 +1,39 @@
...
 # are collected in @updated_meteric_answers and can be executed
-def refresh_answer_lookup_entry answer_id
+def update_answer id: nil, answer_id: nil, metric_id: nil
+  ids = if metric_id
+          Answer.where(metric_id: metric_id).pluck :id
+        elsif answer_id
+          Answer.where(answer_id: answer_id).pluck :id
+        else
+          [id]
+        end
```
ethn: seems like this should be two or three methods, though it's not obvious to me how best to do it  maybe:
```
def update_answers *filters
  ids = answer_ids_to_update *filters
  update_answers_now_or_later ids
end
``` 
?

mod/metric_answer_lookup/lib/answer/cardless_answers.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def virtual_answer_card name=nil, val=nil
+      name ||= [record_name, year.to_s]
+      val ||= value
+      Card.new(name: name, type_id: Card::MetricValueID).tap do |card|
+        card.define_singleton_method(:value) { val }
+        # card.define_singleton_method(:updated_at) { updated_at }
+        card.define_singleton_method(:value_card) do
+          Card.new name: [name, :value], content: val
+        end
```
ethn, line 17: dunno if you saw my outdated comment on this:
> we might consider adding all these virtual card methods to a module and having the singleton_class include it. My hunch is that it would be faster (performance-wise)

I don't want to push for that, but I am curious to hear your rationale for this approach.

mod/metric_answer_lookup/lib/answer/cardless_answers.rb
```diff
@@ -0,0 +1,66 @@
...
+      refresh
+      update_cached_counts
+      self
+    end
+
+    def update_cached_counts
+      [[metric_id, :value], [metric_id, :wikirate_company],
+       [company_id, :metric], [company_id, :wikirate_topic]].each do |mark|
+        Card.fetch(mark).update_cached_count
+      end
```
ethn, line 38: this is pretty slick.

mod/metrics/lib/formula/calculator/input_values.rb
```diff
@@ -0,0 +1,225 @@
+module Formula
+  class Calculator
+    # It finds all metric values and yearly values that are related to the
+    # the metrics and yearlys variables of a formula and prepares the values
```
ethn: yearly (typo)

ethn (approved):
Great stuff.  The calculations appear cleaner than they started.  I'm interested in responses to a few comments, but there is nothing here worth slowing down the merge.

mod/badges/spec/support/award_answer_create_badges_shared_examples.rb
xithan: I reduced the number of tests while I was debugging this (since these are shared examples it's not possible to run them one by one)

mod/caching/set/abstract/answer_table_cached_count.rb
xithan: There are two methods involved now that cache their result: `wql_hash` and `fetch_query`. That's why I introduced the `cache_query?` method. Overriding `wql_hash` means we have to duplicate the code that handles filter and sort options.
At the other places where we override wql_hash the result doesn't change (like here where a a search is involved)

mod/metric_answer_lookup/lib/answer/cardless_answers.rb
xithan: the rationale for this approach: less lines of code and better performance
```ruby
class A; end

def with_module val
  mod = Module.new do
    define_method :x do
      puts val
    end
    define_method :y do
      puts val
    end
  end
  a = A.new
  a.singleton_class.include mod
  a.x
end

def with_method val
  a = A.new
  a.define_singleton_method :x do
    puts val
  end
  a.define_singleton_method :y do
    puts val
  end
  a.x
end
```
Comparison:
         with method:    51961.0 i/s
         with module:    37984.6 i/s - 1.37x  slower

ethn: But in the test you're defining the module over and over again; that's definitely going to be slower.  I was suggesting that you define the module once and then include it over and over again.  I don't _really_ know, but I'd predict that would be an order of magnitude faster.

xithan: I don't see how I can define the module only once. The definition of the module depends on the argument of the method that includes the module. That's what I simulated in my test.

xithan: ok, I could do that with instance variables like this
```ruby
class A; end

module M
  attr_accessor :val1, :val2
  def x; puts val1 end
  def y; puts val2 end
end

def with_single_module val1, val2
  a = A.new
  a.singleton_class.include M
  a.val1 = val1
  a.val2 = val2
  a.x
  a.y
end
```
but it makes no difference:

Comparison:
methods defined at runtime:    60403.9 i/s
hard-coded module included at runtime:    59385.8 i/s - same-ish: difference falls within error
module defined at runtime:    49653.3 i/s - 1.22x  slower

xithan: the only way I found so far to make it faster is if the module is initially prepended: 
```ruby
class A
  def x
    puts "original method"
  end
end

module BM
  attr_accessor :val1, :val2
  def x
    return super unless val1
    puts val1
  end
end


class B < A
  prepend BM
end
```
Including the module at runtime is 1.25x slower than using `B` like this
```ruby
a = B.new
a.val1 = val1 # change method x by setting instance variable
a.x
```

xithan: if you want to try yourself: https://gist.github.com/xithan/dd6b77c8ca88b99e9f27da21cd37b1e2

ethn: OK, I get it now and see why you've done it this way.  Sorry, should have studied a bit closer.  Thank you!

PR #741, opened 2017-12-22
Ux improvements

srivig:
Fixes:
- Company details icon bug 
- Vote button 
- handle "unknown" and "invalid" score values

### DIFF
--- a/mod/metrics/set/abstract/metric_answer/views.rb
+++ b/mod/metrics/set/abstract/metric_answer/views.rb
@@ -31,9 +31,10 @@ format :html do
   end
 
   view :sources do
+    source_card = card.fetch trait: :source
     output [
-      wrap_with(:h5, "Cited"),
-      subformat(card.fetch(trait: :source)).render_core(items: { view: :cited })
+      wrap_with(:h5, "Citations (#{source_card.item_names.size})"),
+      nest(source_card, view: :core, items: { view: :cited })
     ]
   end
 
--- a/mod/metric_answer_lookup/lib/answer.rb
+++ b/mod/metric_answer_lookup/lib/answer.rb
@@ -54,6 +54,10 @@ class Answer < ApplicationRecord
     metric_name.to_name.key
   end
 
+  def updater_id
+    editor_id || creator_id
+  end
+
   private
 
   def ensure_record metric_card, company
--- a/mod/metric_answer_lookup/lib/answer/entry_fetch.rb
+++ b/mod/metric_answer_lookup/lib/answer/entry_fetch.rb
@@ -51,7 +51,7 @@ class Answer
     end
 
     def fetch_editor_id
-      card.updater_id if card.updated_at && card.updated_at > card.created_at
+      card.value_card.updater_id if value_updated?
     end
 
     def fetch_designer_name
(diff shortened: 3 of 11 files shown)


mod/metrics/set/abstract/metric_answer/views.rb
```diff
@@ -60,7 +60,13 @@ def default_value_link_args _args
   end
 
   def beautify value
-    card.scored? ? colorify(value) : value
+    card.scored? ? beautify_scored(value) : value
+  end
+
+  def beautify_scored value
+   no_number = Integer(value) rescue false
```
ethn: I think by the time beautify is called, the value argument may already have been altered by `humanized_value`.

mod/metrics/set/abstract/metric_answer/views.rb
```diff
@@ -60,7 +60,13 @@ def default_value_link_args _args
   end
 
   def beautify value
-    card.scored? ? colorify(value) : value
+    card.scored? ? beautify_scored(value) : value
```
ethn: this is not your fault, but we should be using our set module pattern here.  everything that applies only to scores should be in metric_type/score.rb.

Maybe I'll take that on...

mod/metrics/set/all/colours.rb
```diff
@@ -35,9 +35,18 @@ def mono_colour
     ]
   end
 
+  # Colours for non-numbers
+  def light_colour
+    [
+      "#cccccc",
+      "#e4e9f1"
+    ]
```
ethn: this can be on one line.

mod/metrics/set/abstract/metric_answer/views.rb
```diff
@@ -60,7 +60,13 @@ def default_value_link_args _args
...
+  end
+
+  def beautify_scored value
+   no_number = Integer(value) rescue false
+   value = value.to_sym == :Unknown ? "?" : "!" unless no_number
```
ethn: we shouldn't have to do this test a bunch.  

card.numeric_value? needs to handle invalid.

I would also note that it's quite confusing that `no_number` is true if it's a number and false if it's not.

I want to rearrange things a bit.  If you can get codeclimate and semaphore passing, I'll tackle the needed code improvements.

mod/metrics/set/all/colours.rb
```diff
@@ -35,12 +35,24 @@ def mono_colour
     ]
   end
 
+  # Colours for non-numbers
+  def light_colour
```
xithan: Everywhere else including css it's `color` not `colour`

mod/metrics/set/all/colours.rb
```diff
@@ -35,12 +35,24 @@ def mono_colour
...
+  end
+
   def colorify value, mono=false
     return "" if value == ""
-    colour = pick_colour(value, mono)
+    colour = numeric?(value) ? pick_colour(value, mono) : light_colour[0]
     wrap_color_div(value, colour)
   end
 
+  def numeric? value
```
xithan: There is already a `number?` methods for strings

xithan (changes requested):
Isn't it a cleaner solution to use css classes (like `mono-color-gradient-1`, `multi-color-gradient-5`) instead of inline styles?

mod/metrics/set/all/colours.rb
srivig: Ah, I wasn't aware

mod/metrics/set/all/colours.rb
srivig: Shouldn't we use British English within code :D ?

ethn (approved):

xithan (approved):

PR #746, opened 2018-01-11
Filter

ethn:
fixing bugs exposed by calculated metrics' use of filters.


mod/filter/set/right/browse_metric_filter.rb
```diff
@@ -55,27 +55,20 @@ def sort_options
...
-    metric_type_select
-  end
-
-  view :research_policy_formgroup, cache: :never do
-    research_policy_select
-  end
-
-  view :wikirate_topic_formgroup, cache: :never do
-    autocomplete_filter :wikirate_topic
-  end
```
xithan, line 68: Is this correct that we want different types of input fields for these?

xithan (approved):
in right_filter_form.rb there is more wikirate specific code in decko

mod/filter/set/right/browse_metric_filter.rb
ethn: Well, it certainly looks like the browse metric filter was _trying_ to override the default filter formgroups.  But these were using the wrong view names for that (they would need to start with _filter_.  I'm not actually done with this PR yet, but most of the requests I've been getting are to standardize the interface (for topics, research policy, etc).  If I go back to overriding, I'll document the rationale.

PR #749, opened 2018-01-16
Answer.unknown?

ethn:
more standardized handling of "unknown" values.


mod/metrics/set/abstract/metric_answer/validate.rb
```diff
@@ -3,32 +3,39 @@
 end
 
 event :validate_value_type, :validate, on: :save, when: :standard? do
```
ethn, line 5: I'm realizing that codeclimate isn't being very critical of complex blocks in event/view definitions.  I suspect that helps explain why it has generally seemed less critical of wikirate than decko...

script/standardize_metric_values.rb
```diff
@@ -72,21 +72,21 @@ def convert_potential_metrics potential_result
...
+    else
+      skip_metric m
     end
     puts "======== Finished normalizing #{m.name} ========".blue
   end
 end
 
-def unknown_inside? metric_values
+def valid_values? metric_values
   metric_values.each do |mv|
```
xithan: better: `metric_values.all?`

db/migrate_cards/20180105134604_created_at_for_answers.rb
```diff
@@ -3,7 +3,12 @@
 class CreatedAtForAnswers < Card::Migration
   def up
     Answer.where(created_at: nil).each do |a|
-      a.update_attributes! created_at: a.card.created_at
+      next unless a.card
+      if a.card.type_id == Card::MetricValueID
+        a.update_attributes! created_at: a.card.created_at
+      else
+        a.destroy!
```
xithan: we could use this to get rid of the calculated cards

mod/metrics/set/right/variables.rb
```diff
@@ -2,6 +2,11 @@
 include_set Abstract::Variable
 include_set Abstract::Table
 
+# FIXME: following does not prevent storage.
+event :abort_storage, :validate, on: :save do
+  abort :success
+end
```
xithan: remove it if it doesn't do anything?

xithan (approved):

mod/metrics/set/right/variables.rb
ethn: commenting for now.  The story is still open on PT and we'll need to revisit it fairly soon.

db/migrate_cards/20180105134605_created_at_for_answers.rb
```diff
@@ -0,0 +1,15 @@
+# -*- encoding : utf-8 -*-
+
+class CreatedAtForAnswers < Card::Migration
+  def up
+    Answer.where(created_at: nil).each do |a|
+      next unless a.card
+      if a.card.type_id == Card::MetricValueID && a.card.researched?
+        a.update_attributes! created_at: a.card.created_at
+      else
+        a.destroy!
```
xithan: This needs a little tweak for the case that the answer is calculated
1) we want to keep the answer
2) we want to add the created_at attribute (otherwise that information is lost forever)
3) we have to set `a.answer_id` to nil (to mark it as an answer without card)
4) we can't call `delete!` on the card. I expect there is an event somewhere that removes the answer if we do that.

xithan (changes requested):

db/migrate_cards/20180105134605_created_at_for_answers.rb
ethn: OK, good call. I think my new solution handles these. I think 1-3 is obvious. I found the event for #4 and initially altered it, but then I realized that the delete does the lookup via answer_id, so when the answer_id is nil, it won't delete the Answer.

db/migrate_cards/20180105134605_created_at_for_answers.rb
```diff
@@ -0,0 +1,19 @@
+# -*- encoding : utf-8 -*-
+
+class CreatedAtForAnswers < Card::Migration
+  def up
+    Answer.where(created_at: nil).each do |a|
+      next unless a.card
+      if a.card.type_id == Card::MetricValueID
+        if a.card.calculated?
+          a.update_attributes! created_at: a.card.created_at, answer_id: nil
+          a.card.delete!
```
ethn: I think this a.card.delete! is safe, because the event that deletes the lookup record uses answer_id (and that is now nil).

mod/metric_answer_lookup/set/type_plus_right/metric_value/value.rb
```diff
@@ -3,6 +3,7 @@
 end
 
 event :delete_answer_lookup_table_entry_due_to_value_change, :finalize, on: :delete do
+  # return if metric_card.calculated?
```
ethn, line 6: this is the event that deletes the answer lookup record.

db/migrate_cards/20180105134605_created_at_for_answers.rb
```diff
@@ -0,0 +1,19 @@
...
+class CreatedAtForAnswers < Card::Migration
+  def up
+    Answer.where(created_at: nil).each do |a|
+      next unless a.card
+      if a.card.type_id == Card::MetricValueID
+        if a.card.researched?
+          a.update_attributes! created_at: a.card.created_at
+        else
+          a.update_attributes! created_at: a.card.created_at, answer_id: nil
+          a.card.delete!
```
ethn, line 12: sigh, I keep losing this comment because of tweaks:

this `delete!` should not cause the answer lookup record to be deleted, because that event uses the answer_id to find the answer (and the answer_id is now nil).

ethn: so I think this addresses all four of your (valid) concerns.

xithan (approved):
👍

PR #751, opened 2018-01-22
Staging

ethn:


db/migrate_cards/20180105134606_created_at_for_answers.rb
```diff
@@ -0,0 +1,26 @@
...
+            a.update_attributes! created_at: a.card.created_at
+          else
+            a.created_at = a.card.created_at
+            a.answer_id = nil
+            a.card.delete!
+            a.save!
+            puts "successfully migrated calculated answer: #{a.id}"
+          end
+        rescue
+          puts "problem migrating answer: #{a.id} "
```
ethn, line 19: so this basically didn't work.  the cards didn't delete and pretty much every calculated value hit this rescue case.

I didn't think we could hold up the deployment for this, but it's still unresolved.

ethn (commented):
I don't love the changes in this PR, but we need to merge this in for master to be consistent with what's been deployed

xithan (approved):

PR #752, opened 2018-01-24
research page

xithan:


db/migrate_cards/20171110164955_customizable_skin.rb
```diff
@@ -0,0 +1,85 @@
...
+
+class CustomizableSkin < Card::Migration::Core
+  def up
+    data_dir = File.expand_path("../data/custom_theme", __FILE__)
+    skin = CustomSkin.new("customizable bootstrap", data_dir)
+    skin.create_or_update
+  end
+
+  class CustomSkin
+    THEME_FIELDS = %w[colors components spacing cards fonts more]
```
ethn, line 11: wow!

mod/Modfile
```diff
@@ -2,6 +2,7 @@ mod "wikirate_utility"
 mod "filter"
 mod "layout"
 mod "metrics"
+mod "research"
```
ethn, line 5: 👍

mod/research/set/self/research_page/slot_machine.haml
```diff
@@ -0,0 +1,134 @@
...
+          %h5
+            Question
+          .large-text
+            = nest [metric, :question], view: :core
+        .p-2.pl-3
+          = answer_slot
+
+
+    .col-7.border-left#research-details
+      = right_side_tabs
```
ethn: record for longest WikiRate haml file = crushed.

I'm not sure there's a good reason to break it up, though.

ethn (approved):
Tons of great stuff. Really fun to review.

noticing lots of `tags: :unknown_ok`. Wondering if we should review that API. It's not quite pretty enough for something we have to use all the time (should we consider reversing the default?)

mod/research/set/self/research_page/slot_machine.haml
xithan: there is a lot of repetition but also a lot of little differences

db/migrate_cards/20171110164955_customizable_skin.rb
xithan: This is old code. It's here for the bootstrap upgrade. There is a bug in the migrations. Migrations in core mods don't run. I didn't want to deal with that now so I copied the migration to wikirate.

mod/metrics/lib/javascript/script_answer_source_handling.js.coffee
```diff
@@ -75,65 +73,79 @@ $.extend wikirate,
...
+#
+#    if (!$loader.isLoading())
+#      $('._blank_state_message').remove()
+#      $('.source-details-toggle').removeClass('active')
+#      $(sourceSelector + '.source-details-toggle').addClass('active')
+#      # $this.addClass("active")
+#      $source_form_container.find('.source-details').addClass('hide')
+#      $sourcePreview = $source_form_container.find(sourceSelector)
+#      $source_form_container.find('form').addClass('hide')
+#      wikirate.handleYearData($parentForm, sourceYear)
```
xithan, line 146: I have no idea what this `handleYearData` did.

PR #754, opened 2018-01-26
research page fixes

xithan:


mod/metrics/set/abstract/metric_answer/research_editor.rb
```diff
@@ -11,7 +11,7 @@
   end
 
   def research_form_success
-    { id: "_self", soft_redirect: true, view: :titled }
+    { id: "_self", soft_redirect: true, view: :titled, slot: { title: "Answer" } }
```
ethn, line 14: fine, but are we losing the title from the slot at some point?

mod/metrics/set/abstract/value/edit.rb
```diff
@@ -38,8 +38,8 @@ def editor
   def edit_fields
     return if voo.editor == :standard
     [
-      [card, { title: "Answer", editor: :standard }],
-      [unknown_field_card, { hide: :title }],
+      [card, { title: "Answer", editor: :standard, hide: :help }],
```
ethn, line 41: oh, here we're using "Answer" as the title of the value card, and above we're using it as the title of the whole answer?

My developer brain says the Answer should be the Answer and the Value the value, but having both answer and value is probably confusing, so I guess this is best.

...but my developer brain needed to talk to another developer brain ;)

mod/research/set/self/research_page.rb
```diff
@@ -55,11 +55,12 @@ def right_side_tabs
     tabs = {}
     if answer?
       tabs["Source"] = _render_source_tab
-      tabs["Source preview"] = _render_source_preview_tab
+      tabs["View Source"] = { content: _render_source_preview_tab,
+                              button_attr: { class: "d-none" } }
     end
```
ethn, line 66: So we don't have these tabs at all until there's an answer, is that correct?

mod/research/set/type/source.rb
```diff
@@ -1,18 +1,19 @@
...
+                           "data-form-for" => "new_metric_value",
+                           class: "slotter new-view TYPE-source" do
+          output [
+            new_research_hidden,
+            new_view_type,
+            haml(:source_form),
+            new_research_buttons
+          ]
+        end
       end
```
ethn: extra indention?

I guess I don't mind if code climate doesn't.

mod/research/set/type/source/source_form.haml
```diff
@@ -43,5 +44,5 @@
         = field_nest :wikirate_topic, view: :titled,  title: "Topics", items: { view: :link }
         = field_nest :date, view: :titled, title: "Date"
         = field_nest :wikirate_website, view: :titled, title: "Website", items: { view: :link }
-  %div{:style => "display:none;"}
-    = field_nest :source_type
+  .d-none
```
ethn: +1

mod/wikirate_source/lib/javascript/script_new_source_page.js.coffee
```diff
@@ -1,14 +1,25 @@
...
-      source_type_subcard.siblings("ul").find("input[value=" + source_type + "]").prop("checked", true)
-
 $(window).ready ->
+  $('body').on 'shown.bs.tab', 'form.TYPE-source.new-view .new-source-widget a[data-toggle="tab"]', (e) ->
+    source_type = $(e.target).data('source-type')
+    source_type_subcard = $(e.target).closest("form").find(".RIGHT-Xsource_type input:hidden.d0-card-content")
+
+    source_type_subcard.val("[[" + source_type + "]]")
+    source_type_subcard.siblings("ul").find("input[value=" + source_type + "]").prop("checked", true)
+
```
ethn: as a general pattern, I'm thinking we might want to put these kinds of things into functions.  Mostly because that would make this easier to read.  

If we don't have that, we should probably at least put a comment, eg before the `on` trigger.

mod/wikirate_source/set/type/source/source_item.rb
```diff
@@ -156,12 +155,13 @@ def hidden_item_input
   end
 
   view :source_and_preview, cache: :never do |args|
-    wrap_with :div, class: "source-details",
-                    data: { source_for: card.name, year: year } do
-      args[:url] = source_url
-      with_cite_button +
-        render_iframe_view(args).html_safe +
+    args[:url] = source_url
```
ethn, line 158: fwiw, doesn't look to me like either of those views does anything with args[:url]

mod/metrics/set/type_plus_right/metric_value/checked_by.rb
```diff
@@ -1,263 +1,266 @@
...
+#   !add_checked_flag? && !remove_checked_flag?
+# end
+#
+# format :json do
+#   view :essential do
+#     {
+#       checks: card.checkers.count,
+#       check_requested: card.check_requested?
+#     }
+#   end
```
ethn: We can safely delete this, no?

ethn (approved):

mod/research/set/self/research_page.rb
xithan: that's correct

PR #761, opened 2018-02-06
fix relationship and calculated metrics bugs

xithan:


mod/metrics/set/abstract/04_chart.rb
```diff
@@ -76,7 +76,7 @@ def chart_load_url
   end
 
   def show_chart?
-    return unless card.numeric? || card.categorical?
+    return if !card.relationship? && !card.numeric? && !card.categorical?
```
ethn: `if !A && !B && !C` is better than `unless A || B || C`?

mod/metrics/set/ltype_rtype/metric/wikirate_company/record.rb
```diff
@@ -43,7 +43,7 @@ def more_buttons
     nest metric_card, opts
   end
 
-  view :answer_table do
+  view :answer_table, cache: :never do
```
ethn: why can't we cache this?

mod/metrics/set/type/metric_value/views.rb
```diff
@@ -78,7 +78,7 @@ def year_and_value
     HTML
   end
   # value, unit, and flags
-  view :value do
+  view :value, cache: :never do
```
ethn: why can't we cache this?

ethn (approved):
I'm still worried about overuse of `cache: :never`. Any time you can leave comments in the code about why it's necessary, it will be helpful for when I can go back to try to improve the caching system.

PR #762, opened 2018-02-16
decko tick to fix slot issues

xithan:
also related to the bug that year changes for answer don't work


mod/metrics/set/abstract/metric_answer/validate.rb
```diff
@@ -44,9 +44,10 @@ def validate_category_value value
   end
   self.name = new_name
   detach_subfield(:year)
+  success.year = new_year if success.year
```
ethn, line 47: this makes it look like the year was getting saved correctly but the success wasn't showing it (?).

Doesn't seem like that's the whole problem, right?

ethn (approved):

mod/metrics/set/abstract/metric_answer/validate.rb
xithan: there are three problems:
1. the form sends the wrong year because javascript fails before it reaches the method that updates the hidden input content tags (I fixed that about a week ago but it hasn't been merged and deployed)
2. wagn hitch because the index for card keys saves only 64 characters but the last digit of the year in the example is the 65th character. So the change of the year can't be saved because the key (truncated to 64 characters) is not unique
3. the research page doesn't switch to the changed year

PR #764, opened 2018-02-18
Jspath

ethn:
This updates to use the new API and reactivates the filtered list editor for projects+companies.  The actual filtered list solution is in decko.


xithan:
So the filtered list solution is to use ids instead of names? The average id has 6.6 digits, the average company name has 21 characters. That's an increase by 3.5. The list was broken with about 250 companies. That means it will break again with 900 entries.

ethn:
I know.  See the discussion in here: https://www.pivotaltracker.com/story/show/155215791

I had estimated 800, though not as carefully as you.  If we wanted to be precise, we'd account for the fact that I took out the brackets (6 characters when CGI escaped), left in the line breaks (3 when escaped), and of course there's the rest of the URL...

More to the point, this is a one-line fix that let's us turn things back on.  The deeper fix means implementing something a bit heavier.

I'd like to go ahead and merge this and deploy it along with the rest of last week's fixes tonight.

xithan (approved):

PR #766, opened 2018-02-21
hybrid metrics

xithan:


db/migrate_cards/20180213161036_add_hybrid_flag.rb
```diff
@@ -0,0 +1,10 @@
+# -*- encoding : utf-8 -*-
+
+class AddHybridFlag < Card::Migration
+  def up
+    ensure_trait "hybrid", :hybrid,
+                 default: { type_id: Card::ToggleID },
+                 help: "Allow calculated answers to be overridden based on sourced "\
+                       "research."
```
ethn: Slight change (as discussed in ticket):

> Allow answers to be directly added or overridden based on sourced research

mod/metric_answer_lookup/lib/answer/cardless_answers.rb
```diff
@@ -6,6 +6,10 @@ def self.included host_class
       host_class.extend ClassMethods
     end
 
+    def find_answer_card
+      Card[record_name, year.to_s, :value]&.left
```
ethn: maybe add comment explaining why we get the value and then left (rather than going straight to the answer card)

db/migrate_cards/20180213161036_add_hybrid_flag.rb
ethn: we can add an answer if there's not a calculated answer, right?

mod/metric_answer_lookup/lib/lookup_table.rb
```diff
@@ -4,7 +4,9 @@ def card
   end
 
   def refresh *fields
-    return delete if !card || card.trash || invalid?
+    # when we override a hybrid metric the answer is invalid because of the
+    # missing answer_id, so we check `invalid?` only for non-hybrid metrics)
```
ethn, line 8: do I understand correctly that this means the answer_id is missing _before_ we override (but the answer is there once overridden?) Or are we not storing answer_ids when we override?

mod/metric_answer_lookup/set/type/metric_value.rb
```diff
@@ -1,3 +1,3 @@
 event :update_answer_lookup_table_due_to_answer_change, :finalize, on: :update do
-  update_answer answer_id: id
+  hybrid? ? update_answer(id: answer.id) : update_answer(answer_id: id)
```
ethn, line 2: `update_answer(id: answer.id)`   is this just to tick updated_at or something?

db/migrate_cards/20180213161036_add_hybrid_flag.rb
ethn: nevermind. the cuke answers this question.

mod/metrics/set/abstract/metric_answer/source_citation.rb
```diff
@@ -14,6 +14,10 @@
   end
```
ethn, line 14: Do I understand correctly that we never get here for calculated answers (because they're not created)?

ethn (approved):
Seems like we're going to want to let folks play with this on staging for a while before we deploy. Let's get this and research groups ready for them to test by the end of the week.

Looks like very solid stuff.

mod/metric_answer_lookup/set/type/metric_value.rb
xithan: since calculated answers don't have an answer_id we can't find them using the answer_id

mod/metrics/set/abstract/metric_answer/source_citation.rb
xithan: We never get here because  `source_based?` is false for calculated answers

mod/metric_answer_lookup/lib/lookup_table.rb
xithan: the former is correct

mod/metric_answer_lookup/set/type/metric_value.rb
ethn: oh, I misread this and was thinking we were updating the attribute.  doh.

PR #770, opened 2018-02-28
update json api for ccc widget

xithan:


mod/profiles/set/self/wikirate_company.rb
```diff
@@ -8,12 +8,14 @@ def ids_related_to_research_group research_group
...
       oc_card = Card.fetch(id, :open_corporates)
+      wikipedia_card = Card.quick_fetch(id, :wikipedia)
       comps << {
         id: id,
         name: Card.fetch_name(id),
         alias: (alias_card&.item_names || []),
         oc_id: oc_card&.company_number,
-        oc_jurisdiction_code: oc_card&.jurisdiction_code
+        oc_jurisdiction_code: oc_card&.jurisdiction_code,
+        wikipedia: wikipedia_card&.content
```
xithan: requested by Vasso

mod/widgets/set/self/jurisdiction.rb
```diff
@@ -1,11 +1,63 @@
...
+    h = Hash.new { |hash, key| hash[key] = {} }
+    by_country = Card.search(type_id: JurisdictionID).each_with_object(h) do |i, groups|
+      if (m = i.name.match(/(?<state>.+?)\s*\((?<country>[^)]+)\)/))
+        groups[m[:country]][:text] ||= m[:country]
+        groups[m[:country]][:children] ||= []
+        groups[m[:country]][:children] << { id: i.codename, text: m[:state] }
+      else
+        groups[i.name][:text] ||= i.name
+        groups[i.name][:id] = i.codename
+      end
```
ethn: this would be a good candidate for multiple methods.
```
if (m = i.name.match REGEXPCONSTANT)
  country_details groups[m[:country]], i, m
else
  group_details groups[:i.name], i
end
```

mod/widgets/set/self/jurisdiction.rb
```diff
@@ -1,11 +1,63 @@
...
+      if data.key?(:children)
+        if data[:children].size > 1   # valid group
+          # group header is an item itself
+          result << { id: data.delete(:id), text: data[:text] } if data.key?(:id)
+          result << data
+        else  # just a single item
+          child = data[:children].first
+          child[:text] = "#{child[:text]} (#{key})"
+          result << child
+        end
```
ethn: another candidate.

mod/profiles/set/self/wikirate_company.rb
ethn: wikipedia_title would be slightly more precise

mod/profiles/set/self/wikirate_company/factories.rb
```diff
@@ -4,7 +4,10 @@
     if keyword
       company_ids = search_by_company_name(company_ids) | search_by_address(company_ids)
     end
-    company_ids.map { |id| Card.fetch_name(id) }.to_json
+    company_ids.map do |id|
+      company_name = Card.fetch_name(id)
+      { name: company_name, id: id, url_key: company_name.url_key }
+    end.to_json
```
ethn: separate method?

mod/profiles/set/self/wikirate_company/factories.rb
```diff
@@ -39,10 +42,10 @@ def search_by_country
   end
 
   def country_code
-    Env.params[:country_code]
+    Env.params[:country_code] if Env.params[:country_code].present?
   end
 
   def keyword
-    Env.params[:keyword]
+    Env.params[:keyword] if Env.params[:keyword].present?
```
ethn: would be nice to have a shortcut for this.  But you can just use `param` in format objects.  (see def params in format.rb)

ethn (approved):

mod/profiles/set/self/wikirate_company/factories.rb
xithan: I always forget if it's defined in format or card objects so I always use `Env` to be safe.

mod/widgets/set/self/jurisdiction.rb
xithan: hm, I don't like these kind of side effect methods

PR #771, opened 2018-03-01
ccc api fix

xithan:


mod/profiles/set/self/wikirate_company/factories.rb
```diff
@@ -1,32 +1,32 @@
 format :json do
   view :search_factories, cache: :never do
-    company_ids = search_by_country
-    if keyword
-      company_ids = search_by_company_name(company_ids) | search_by_address(company_ids)
-    end
-    company_traits(company_ids).to_json
+    ids = if keyword
+            search_by_company_name | search_by_address
```
ethn, line 4: Let's say I had a company "Big Pants" at "89 Cuckoo Lane".  If I searched for "Pants Cuckoo", it wouldn't find it, right?  Because it would search for companies with both and addresses with both.

Either we need to change the help text to "company or address" or we need to come up with far more sophisticated matching.

ethn: (I know this PR doesn't introduce that issue; it was my concern from the start but I didn't articulate it well before).

ethn (approved):

mod/profiles/set/self/wikirate_company/factories.rb
xithan: Doen't "/" mean or?

ethn: I read it as "and/or".

PR #774, opened 2018-03-05
first go at box view (for vignesh)

ethn:
I think we can merge this now.

It basically introduces the boxes API, but it's not really used yet.  We're not scheduled to dive all the way in on this for a while, so I'm hoping to merge in what we've done and move on, provided we don't break anything.


mod/metrics/set/type/metric/structure.rb
```diff
@@ -35,14 +37,21 @@ def right_column_class
   end
 
   view :metric_question do
-    question = subformat(card.question_card)._render_content
     wrap_with :div, question, class: "question blockquote"
   end
 
+  def question
+    subformat(card.question_card)._render_content
```
ethn: if all we want is the plain text of the question, `card.question` will do.

render_content adds handling for nest rendering, in place editing, a div with handy css classes.  Do we want that here?

ethn: also, standard pattern is `nest(card.question_card, view: :content)`

mod/metrics/set/type/metric/structure.rb
```diff
@@ -35,14 +37,21 @@ def right_column_class
...
+  end
+
   view :designer_info do
     nest card.metric_designer_card, view: :designer_info
   end
 
+  view :designer_info_without_label do
+    nest card.metric_designer_card, view: :designer_info_without_label
```
ethn: we should consider using `hide: designer_label`

xithan (approved):

PR #785, opened 2018-03-21
 fixed metric counts on topic+company and company+topic

ethn:
...and topic count on homepage

### DIFF
--- a/mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
+++ b/mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
@@ -24,11 +24,9 @@ end
 def wql_from_content
   metric_ids = unique_metric_ids
   if metric_ids.any?
-    {
-      type_id: WikirateTopicID,
+    { type_id: WikirateTopicID,
       referred_to_by: { left_id: [:in] + metric_ids,
-                        right_id: WikirateTopicID }
-    }
+                        right_id: WikirateTopicID } }
   else
     { id: -1 } # HACK: ensure no results
   end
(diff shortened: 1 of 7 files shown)


mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
```diff
@@ -45,13 +43,28 @@ def unique_metric_ids
...
 
+# @returns Array [[topic_card1, num_metrics], [topic_card2...]...] }
 def topics_by_metric_count
-  topic_count_hash =
-    item_cards(limit: 0).each_with_object({}) do |topic_card, count_hash|
-      count_hash[topic_card] = topic_card.metric_card.cached_count
-      count_hash
-    end
-  topic_count_hash.sort_by { |_card, count| count }.reverse
+  topic_metric_count_hash.sort_by { |_card, count| count }.reverse
```
xithan, line 48: You didn't introduce it but `sort_by` is built for the case that key comparison is very expensive 
(e.g. every comparison creates a complicated object). In this case where comparison is simple the `sort` method is much faster.

xithan (changes requested):
has only a simple performance issue

mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
ethn: I don't think sort will cut it here.  We want to sort by the value.

```
 {a: 1, b: 10, c: 5 }.sort
=> [[:a, 1], [:b, 10], [:c, 5]]

{a: 1, b: 10, c: 5 }.sort_by { |_k, v| v }
=> [[:a, 1], [:c, 5], [:b, 10]]
```

xithan:
```
{ a: 1, b: 10, c: 5 }.sort { |(k, v), (k1, v1)| v <=> v1 }
=> [[:a, 1], [:c, 5], [:b, 10]]
```

xithan: but it turns out for sorting by hash values `sort_by` is actually faster and looks better:
```ruby
require 'benchmark'

h = (1..1000000).each_with_object({}) { |v, hash| hash[v] = rand(1000000) }
a = (1..1000000).map { rand(1000000) }

Benchmark.bm(10) do |b|
  b.report("h.sort")    { h.sort { |(_k1, v1), (_k2, v2)| v1 <=> v2 } }
  b.report("h.sort_by") { h.sort_by { |_k, v| v } }
  b.report("a.sort")    { a.sort }
  b.report("a.sort_by") { a.sort_by { |a| a } }
end
```
```
                 user     system      total        real
h.sort       3.660000   0.090000   3.750000 (  3.999615)
h.sort_by    1.410000   0.040000   1.450000 (  1.573429)
a.sort       0.240000   0.020000   0.260000 (  0.269645)
a.sort_by    1.260000   0.010000   1.270000 (  1.348517)
```

xithan (approved):

PR #795, opened 2018-04-12
updates to follow decko api changes (head tag, etc)

ethn:


mod/research/set/self/research_page/params.rb
```diff
@@ -74,7 +74,7 @@ def research_param_keys
     if project?
       keys << :project
     else
-      keys += %i[metric_id_list company_id_list year_id_list]
+      keys += %i[metric_id_list company_id_list] # year_id_list]
```
xithan: you can probably remove it if there are no values in year_id_list but removing it always seems like a bad idea to me

xithan (approved):

mod/research/set/self/research_page/params.rb
ethn: I looked for and failed to find a case where this was useful. Can you explain why you'd ever want it?

It creates a lot of noise and makes links all but unsharable.  And it looks like it just gets populated for no good reason.

test/shared_data/relationship_metrics.rb
```diff
@@ -35,14 +34,18 @@ def relationship_metrics
         Monster_Inc "1977" => { "Los_Pollos_Hermanos" => "Tier 1 Supplier" }
       end
 
-      Card::Metric.create name: "Clean Clothes Campaign+Address",
+      address_name = "Clean Clothes Campaign+Address"
```
xithan: `Commons`?

mod/research/set/self/research_page/params.rb
xithan: I think there are additional parameters now for all the cases how we currently access the research page (project or one element pinned) so that we probably don't really need any of these id lists anymore but I don't want to cripple an api in one place and leave code in a lot of other place that doesn't work anymore. I made a PR to your branch to make less noisy urls.

mod/metrics/set/abstract/all_metric_values.rb
```diff
@@ -24,10 +24,13 @@ def query args={}
...
   view :core, cache: :never do
-    class_up "card-slot", "_filter-result-slot"
-    output [_render_filter_form, _render_filter_result]
+    wrap_with :div, class: "filter-form-and-result nodblclick" do
+      class_up "card-slot", "_filter-result-slot"
+      output [_render_filter_form, _render_filter_result]
+    end
   end
-
+gs
```
xithan: 🐛

PR #796, opened 2018-04-12
chart fixes

xithan:
[Finishes #154561495]


mod/metrics/set/abstract/04_chart.rb
```diff
@@ -6,7 +6,13 @@ def filter_hash with_chart_filter=true
...
 
 def chart_params
-  Env.params[:chart].is_a?(Hash) ? Env.params[:chart] : {}
+  if Env.params[:chart].is_a?(Hash)
+    Env.params[:chart]
+  elsif Env.params[:chart].is_a?(ActionController::Parameters)
+    Env.params[:chart].to_unsafe_h
+  else
+    {}
+  end
```
ethn, line 15: seems like we should generalize this in Env (until we work out something more comprehensive)

ethn (approved):

PR #799, opened 2018-04-16
skin fixes

xithan:


db/migrate_cards/20180414102432_move_customized_skin.rb
```diff
@@ -0,0 +1,23 @@
+# -*- encoding : utf-8 -*-
+
+class MoveCustomizedSkin < Card::Migration
```
ethn, line 3: So, do I understand correctly that:
1. most sites don't have a "customized" skin
2. wikirate needs this because it does, and
3. if another site has one, it may struggle with the decko upgrade?

I'm starting to think we label this 0.5 so that we can (fairly) say it was a major upgrade if there are problems :-/

xithan: I wrote the migration that introduced that customizable skin in November 2017. If someone noticed it and used it in that short period then he will struggle with the upgrade

ethn: I'll probably want to warn the Holochain folks.

ethn (approved):

db/migrate_cards/20180414102432_move_customized_skin.rb
xithan: I added a migration for existing customized skins

PR #800, opened 2018-04-18
update to new view api changes

ethn:
not ready for merge


mod/metrics/set/abstract/metric_thumbnail.rb
```diff
@@ -22,11 +22,7 @@ def thumbnail_title_without_link
   end
 
   def thumbnail_title_link
-    opts = { title: card.name }
-    if voo.closest_live_option(:project)
-      opts[:path] = { filter: { project: voo.closest_live_option(:project) } }
-    end
```
ethn, line 28: I couldn't find use of this; it appears to have been replaced with the `metric_links` functionality.

mod/csv_import/set/type_plus_right/source/file/import_answers_from_source.rb
```diff
@@ -58,15 +58,15 @@ def check_card card_name, type_name, type_id
   end
 
   def year_select
+    @explicit_form_prefix = corrections_input_name(:all, :year)
```
ethn, line 61: this is actually the change I'm least comfortable with.  I'm fine with using instance variables, but I think it should be on the format created in the nest.

Any unknown arguments to `Format.new` are turned into instance variables, so I'm thinking that we have nest pass format options to subformat.

More specifically, it would be
```
  def nest cardish, view_options={}, nest_options={}, &block
```
nest_options would include:
- **mode** (currently deleted from opts)
- **comment** (also not a view opt, but it stops the view...)
- **format** (a hash of args to pass to format)

I don't really understand why we need both `mode` and `override`, but if override is necessary, we can put it in there, too.

If we had that, then the current call would be:
```
nest card.left.year_card, { view: :edit_in_form }, format: { explicit_form_prefix: corrections_input_name(:all, :year) }
```
It's not short, but the "nest_opts" are generally seldom used, and I like having the view_opts cleanly limited to the voo options.

mod/metrics/set/abstract/metric_answer/flags.rb
```diff
@@ -27,7 +27,7 @@ def commented?
 
   def checked_value_flag
     return "" unless card.checked?
-    nest card.field(:checked_by), view: :icon, class: "margin-left-10"
```
ethn, line 30: afaict, this was doing nothing.  the icon view didn't handle it.

mod/metrics/set/abstract/metric_answer/source_citation.rb
```diff
@@ -112,9 +112,9 @@ def cited? source_card
 
   def add_source_form
     params[:company] ||= card.company
+    @answer_for_citation = card.name
     nest Card.new(type_id: Card::SourceID),
```
ethn, line 116: this would also use the new api proposed above

mod/metrics/set/abstract/metric_thumbnail.rb
```diff
@@ -37,9 +33,9 @@ def thumbnail_title_text
     "</span>#{_render_thumbnail}<span>"
   end
 
-  view :fixed_value do
-    nest [card, voo.closest_live_option(:params)], view: :value_link
-  end
+  # view :fixed_value do
+  #   nest [card, voo.inherit(:params)], view: :value_link
+  # end
```
ethn, line 38: couldn't find use, including with a WQL search of structure cards on wikirate.org

mod/metrics/template/abstract/metric_answer/_cited_sources.haml
```diff
@@ -1,7 +1,7 @@
 .card-editor.form-group
   - with_nest_mode :normal do
     = field_nest :source, view: :hidden_content_field
-  .cited-sources._pointer-list
+  .cited-sources._pointer-list.pointer-list-ul
```
ethn, line 4: merged in your change from staging

mod/research/set/self/research_page.rb
```diff
@@ -74,8 +75,9 @@ def hide_tab tab, hide=false
 
   view :source_preview_tab, cache: :never do
     wrap do
-      nest preview_source, view: :source_and_preview, cited: cited_preview_source?,
-                           disabled: existing_answer_with_source?
+      @source_cited = cited_preview_source?
+      @source_disabled = existing_answer_with_source?
+      nest preview_source, view: :source_and_preview
```
ethn, line 80: again, would use new proposed api

mod/csv_import/set/type_plus_right/source/file/import_answers_from_source.rb
ethn: by the way, this api means you could nest something with a different format with (`format: { format: :json }`).  Perhaps that's an argument for `format_opts` instead of `format` ?

ethn: ...or all nest_options that nest doesn't know about are forwarded to format...?

So...
```
nest card.left.year_card, { view: :edit_in_form }, explicit_form_prefix: corrections_input_name(:all, :year)
```
I think I like that.

xithan (approved):

PR #805, opened 2018-04-24
new nest and before api

ethn:


mod/metrics/set/abstract/01_researched.rb
```diff
@@ -6,7 +6,7 @@ def score_cards
 end
 
 format :html do
-  def default_content_formgroup_args _args
+  before :content_formgroup do
     super
```
xithan: Does `super` work with the new api? Is it necessary? The new syntax looks more like it adds stuff than that it replaces already existing `before` blocks

ethn: It's actually a pretty tiny change with a fancy dress.

It really replaces the `default_X_args args` with `_before_X` methods (no args).  It's the same concept, but no args, just voo/params/etc.

I evidently didn't merge the "before" stuff into nesting before like I thought I did.  It's here: https://github.com/decko-commons/decko/pull/207/files#diff-f88bb88576209d1b0c98fe0f5c1120c8

xithan: The answer to my question is no, it doesn't work. You need an empty pair of parentheses

xithan (changes requested):

xithan (approved):

PR #810, opened 2018-05-01
fixes post listings and tab icon weirdness

ethn:


mod/wikirate/set/all/wikirate/icons.rb
```diff
@@ -17,6 +17,8 @@
 
 format :html do
   def icon_map key
-    ICON_MAP[key]
+    val = ICON_MAP[key]
+    val[1] = val[1].clone if val.is_a? Array
+    val
```
ethn: clones hashes in ICON_MAP so that it's not broken in args handling.

might be slightly faster further downstream, but this way we're less likely to hit the bug again.

mod/posts/spec/set/type/post_spec.rb
```diff
@@ -1,11 +1,22 @@
...
-        expect_view(view).to lack_errors
+  context "with company, topic, and subject" do
+    # TODO: move this post to seed data
+    let(:post) do
+      Card.create! type: "Post", name: "My Post",
+                   subcards: {
+                     "+company" => "Death Star",
+                     "+topic" => "Force",
+                     "+project" => "Evil Project"
+                   }
```
ethn: tests didn't catch listing view issues because there weren't any items to list.

ethn: (they failed with these changes)

xithan (approved):

PR #813, opened 2018-05-04
version and decko tick.  [Finishes ##157293577]

ethn:
decko is master (with 5.2 fixes) + nest fix.

I'm not sure we want to merge the nest fix into decko until we have tests, but I do think we should deploy to wikirate immediately.


ethn:
I deployed the nest fix (but not the 5.2 fixes) to production

xithan (approved):

ethn:
nothing else needed from here.

PR #816, opened 2018-05-14
scripts and styles clean up

xithan:


mod/homepage/spec/set/self/homepage_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  let(:golden_master) do
+    ::File.read(File.expand_path("../homepage.html", __FILE__))
+  end
+
+  def remove_head html
+    html.gsub(/<head>.+<\/head>/m, "").gsub(/data-card-id="\d+"/, "")
+        .gsub(/src="[^"]+"/, "").strip
+  end
+
+  xspecify "homepage is fine" do
```
ethn: do expect to turn this on soon?

ethn (approved):

mod/homepage/spec/set/self/homepage_spec.rb
xithan: homepage is too broken in the test data. the first step is that the export views on production work again

PR #825, opened 2018-05-24
don't contextualize value options

xithan:


mod/metrics/lib/card/chart/vega_chart.rb
```diff
@@ -282,13 +283,18 @@ def bar_zoom_link_params filter_opts
...
             zoom_out: {
               chart: @format.chart_params,
               filter: @format.filter_hash(false)
             }
           }
         }
       end
+
+      def zoom_level
+        @format.chart_params[:zoom_level].to_i || 0
```
ethn, line 296: ` || 0` might be unnecessary.  `"".to_i` and `nil.to_i` are both 0

mod/metrics/set/abstract/04_chart.rb
```diff
@@ -77,25 +77,26 @@ def vega_chart
...
+    value_filter_to_human
+  end
+
+  def value_filter_to_human
+    if filter_hash[:range]
+      value_range_filter_to_human filter_hash[:range]
     else
-      mv[:numeric_value] || mv[:category]
+      f = filter_hash
+      f[:numeric_value] || f[:category] ||
```
ethn, line 89: I know we can't easily have card sets for value types, but I'm thinking we should make some ruby classes for them.  Their logic is very distributed right now.

ethn (approved):

PR #827, opened 2018-05-25
vote fixes

ethn:
fixed busted link and id representation


xithan:
I think if we turn the vote buttons into links a `rel=nofollow` would be good.

ethn:
nofollow is there now.

xithan (approved):

PR #834, opened 2018-05-30
rename metric value to metric answer

xithan:


db/migrate_cards/20180530085450_rename_metric_value_to_metric_answer.rb
```diff
@@ -0,0 +1,8 @@
+# -*- encoding : utf-8 -*-
+
+class RenameMetricValueToMetricAnswer < Card::Migration
+  def up
+    update_card :metric_value,
+                name: "Metric Answer", codename: "metric_answer"
```
ethn: I think it would be my preference to have the *card name* "Answer" rather than "Metric Answer", because there are places where that copy appears on the site.  (Eg on the contributions tab of user pages).

I do think `metric_answer` makes the most sense as a *codename*, though, both because it avoids current conflicts with the answers table and because it may help avoid future conflicts with other mods that have answers.  (For that reason, I like the practice of prefixing codenames with the modname, even though it makes things longer).

mod/profiles/spec/set/ltype_rtype/user/cardtype_spec.rb
```diff
@@ -1,13 +1,13 @@
 RSpec.describe Card::Set::LtypeRtype::User::Cardtype do
   describe "view :contribution_report" do
     def report_url report
-      "/Joe_User+Metric_value?report_tab=#{report}&view=contribution_report"
+      "/Joe_User+Metric_Answer?report_tab=#{report}&view=contribution_report"
```
ethn: (flag: in case we change card name to Answer)

ethn (changes requested):
very happy.  will be even happier if you agree with me re card name :)

db/migrate_cards/20180530085450_rename_metric_value_to_metric_answer.rb
xithan: On the contributions tab it's still "Metric values". I didn't choose `answer` to avoid conflicts with the answer table but it's probably safe as card name.

ethn: you mean it's still "metric values" even after this change?  (I know it's "metric values" on the live site; I was thinking this would fix that.)

mod/profiles/set/self/metric_answer.rb
```diff
@@ -38,3 +38,7 @@ def created_query_variant user_id, variant=nil
     {}
   end
 end
+
+def contribution_report_title
+  "Metric values"
+end
```
xithan: !

ethn: Confused.  We want the title to be "Answers", right?

(Though this does make the point that we don't have to change the cardname for that).

xithan: I don't know if we want a change in the messaging. This PR never changed anything that is visble by users (to my knowledge). Your very first comment gave the impression that didn't notice that so I highlighted this.

ethn: Now i understand.  You're right, I missed that. I confirmed that we do want "answers"

ethn (approved):

PR #838, opened 2018-06-05
WikiRating

ethn:


mod/metrics/set/right/checked_by.rb
```diff
@@ -76,7 +76,7 @@ def option_names
 end
 
 def answer
-  @answer ||= left&.answer
+  @answer ||= left.answer
```
xithan: are you sure that we don't need the `&`?

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -44,6 +19,104 @@ decko.slotReady (slot) ->
...
+    tr = $(this)
+    if key = tr.find(".metric-label .thumbnail").data "cardName"
+      hash[key] = tr.find(".metric-weight input").val()
+  hash
+
+$(window).ready ->
+  $('body').on 'input', '.metric-weight input', (_event) ->
+    validateWikiRating $(this).closest(".pairs-editor")
+
+  $('body').on "click", "._remove_weight", () ->
```
xithan: `_remove-weight` would be more consistent with our other classes

xithan (approved):

mod/metrics/set/right/checked_by.rb
ethn: I'll put it back.  that was recently added  but didn't solve the problem (solved in `#unknown?`)

PR #840, opened 2018-06-07
show overridden calculated answers

xithan:


mod/research/set/self/research_page/params.rb
```diff
@@ -162,7 +162,7 @@ def researchable_answer?
 
   def existing_answer_with_source?
     existing_answer? && researchable_answer? &&
-      (!answer_card.hybrid? || answer_card.calculation_overridden?)
+      (!answer_card.hybrid? || answer_card.researched_value?)
```
ethn: I find this a little tricky to read, but I spent a few minutes trying to make it easy and basically failed.

db/migrate/20180529102124_add_calculated_value_to_answer_table.rb
```diff
@@ -0,0 +1,7 @@
+# -*- encoding : utf-8 -*-
+
+class AddCalculatedValueToAnswerTable < Card::Migration::DeckStructure
+  def up
+    add_column :answers, :calculated_value, :string
```
ethn: I know it's a long name, but I'm tempted to call this `overridden_calculated_value`, because it took me a bit to figure out that a calculated answer has no calculated_value unless it's overridden.

if we want it short, I think `overridden_value` is less confusing than `calculated_value`

mod/research/set/self/research_page/params.rb
ethn: maybe if the second line were:
`(answer_card.researched? || answer_card.researched_value?)`

ethn (approved):
I do actually request the naming change, but I give you the authority to override me.

db/migrate/20180529102124_add_calculated_value_to_answer_table.rb
xithan: The length of `overridden_calculated_value` was the decisive factor to go for `calculated_value` (but with a comment that describes what it does) . I changed it to `overridden_value` but it's annoying that these little changes tend to take several hours to make them happen. In this case because it involves a card migration and a deck structure migration. I couldn't get the updated structure migration to run again. Eventually I figured out that the migration did run but I didn't notice it because structure migrations don't produce any output when they run (like these messages `== 20180529102142 AddCalculatedValueToAnswerTable: migrating =========` that you usually see) and they don't update the schema.

ethn: Ugh, I'm sorry that took hours. I don't understand why the output isn't working, but the schema issue is the `without_dumping` thing that I added to migrate.rake to avoid the problem of failures attempting to alter files in the gem.  maybe we need to make it smart about whether the decko code is in a gem or a repo.  I can work on that.

PR #850, opened 2018-06-21
[Finishes #158503239] fix record ui wrapping issues

srivig:
I only tested with test db and I still want to test it on real data (my local copy of live db isn't working well)


ethn (approved):
looks good to me. thoughts @xithan?

PR #854, opened 2018-06-24
fixes for duplicates

ethn:


mod/metric_answer_lookup/set/type/metric_answer.rb
```diff
@@ -1,7 +1,12 @@
+
+def answer
+  @answer ||= Answer.existing(id) || Answer.new
+end
```
ethn, line 4: just seems like this belongs here.

xithan (approved):

PR #857, opened 2018-06-29
update docs for new developer orientation

ethn:


README.md
```diff
@@ -1,25 +1,42 @@
...
-4. bundle exec wagn s
+The following will help set up a functioning wikirate site with a small subset of (mostly fake) data.  Some pages will not look complete.
+
+1. clone repo: `git clone git@github.com:wikirate/wikirate.git`
+1. enter dir: `cd wikirate`
+1. init/update submodules `git submodule update -f --init --recursive`
+1. install gems: `bundle install`
+1. set up config: `cp -R sample_config/* config`
+1. populate dev database (with test data): `DATABASE_NAME_TEST=wikirate_dev bundle exec rake wikirate:test:seed`
+1. add AWS credentials to config/application.rb (Ask wikirate dev team!  Sorry, we'll make this easier soon)
```
ethn, line 15: I'm not in love with this step, but it should be fine for Monday.

I'm thinking I should update our credentials and give each of us our own.  Also, I think we should probably make it so that none of our dev credentials empower us to alter anything in the live bucket.

README.md
```diff
@@ -32,34 +49,6 @@ Paths:
  mod/*/features/step_definitions  # cucumber step definitions
 ```
 
-Updating Test DB
-----
-Run `rake wikirate:test:reseed_data` to prepare test database.
```
ethn, line 37: this looked mostly wrong to me (I don't think the initial command even exists).  Is there anything here we need to preserve?

README.md
xithan: I use `development: *test` in my database.yml instead of seeding the dev db with test data.

README.md
xithan: that looks outdated to me

xithan (approved):

README.md
ethn: I often do that, too. This seems like a slightly safer way to start off, though, before we've explained how our test db works.

xithan: I'm using the user "reader" in my dev credentials. I don't think I can alter anything.

PR #858, opened 2018-07-02
expanded relationship answer fixes.

ethn:


test/shared_data/projects.rb
```diff
@@ -0,0 +1,31 @@
...
+                     wikirate_status: "Active",
+                     organizer: "Jedi"
+
+      create_project "Empty Project", metric: [], wikirate_company: []
+    end
+
+    def create_project name, subfields
+      hash = { type: :project, subfields: {} }
+      subfields.each do |codename, values|
+        puts "hash[:subfields][#{codename}] = #{project_subfield values}"
```
xithan: Do you want to keep that `puts`?

test/shared_data/projects.rb
```diff
@@ -0,0 +1,31 @@
...
+        hash[:subfields][codename] = project_subfield values
+      end
+      create name, hash
+    end
+
+    def project_subfield values
+      {
+        type: :pointer,
+        content: (Array.wrap(values).map { |v| "[[#{v}]]" }.join("\n"))
```
xithan: `to_pointer_content` does the same but you don't need it. `content: values` should work.

xithan (approved):

PR #862, opened 2018-07-06
calculated and overridden metric value flag

srivig:
I couldn't test this properly due to issues with my local machine cache. If possible could you test this one your machine?


mod/metrics/set/abstract/metric_answer/flags.rb
```diff
@@ -25,6 +25,21 @@ def commented?
     end
   end
 
+  def calculated_flag
+    return "" unless card.calculated?
+    return hybrid_flag if card.hybrid?
```
ethn: this is a metric method. Any researchable calculated metric is a hybrid.

You want `calculation_overridden?`

Let's not do this second `return`
Try:
```
if A
  B
else
  C
end
```

ethn (changes requested):

mod/metrics/set/abstract/metric_answer/flags.rb
```diff
@@ -25,6 +25,26 @@ def commented?
     end
   end
 
+  def calculated_flag
+    return "" unless card.calculated?
+    calculated_flag_icon
+  end
+
+  def calculated_flag_icon
+    return overridden_flag_icon if card.calculation_overridden?
```
ethn: let's go with `researched_value?` instead of `calculation_overridden?`.  This will handle cases where there is no calculated value to override (because the variable metrics aren't populated).

ethn (changes requested):
I think we can merge this in with the minor change suggested.

ethn (approved):

PR #864, opened 2018-07-17
jest setup

xithan:
adds jest (javascript testing library)
to initialize it get yard if you don't have it and run `yarn install`
to run the tests `jest test`


mod/wikirate/spec/lib/javascript/script_wikirate_common.test.coffee
```diff
@@ -0,0 +1,8 @@
+
+require('../../../lib/javascript/script_wikirate_common.js.coffee')
+
+test 'isString returns true for string', ->
+  expect($.wikirate.isString("ab")).toBe true
+
+test 'isString returns false for integer', ->
+  expect($.wikirate.isString(5)).toBe false
```
xithan: the first and so far only jest tests

README.md
```diff
@@ -36,7 +36,7 @@ Testing
 Running Tests
 ----
 1. populate test database: `bundle exec rake wikirate:test:seed`
-1. start rspec with `decko rspec` and cucumber with `decko cucumber`.
+2. start rspec with `decko rspec`cucumber with `decko cucumber`.
```
ethn: probably want the "and" back.

(fwiw, markdown handles ordered lists with _any_ number):

```
1. 1
5. 2
1. 3
```
=
1. 1
5. 2
1. 3

.pryrc
```diff
@@ -0,0 +1,134 @@
+def create name='test card', content='some content', type='basic'
```
ethn, line 3: As we're adding developers, we probably want to up our commenting game.  For the top section, we don't necessarily need a comment for each method, but maybe something like:
> These commands are available in the console when using `binding.pry` for breakpoints.

mod/wikirate/spec/lib/javascript/script_wikirate_common.test.coffee
```diff
@@ -0,0 +1,8 @@
+
+require('../../../lib/javascript/script_wikirate_common.js.coffee')
+
+test 'isString returns true for string', =>
+  expect($.wikirate.isString("ab")).toBeTruthy()
```
ethn, line 5: I asked in a different PR (from Daniel) whether we wanted a convention for boolean functions in JS, since we can't use question marks.  I'm ok with `isX` and `hasX`.  are you?

ethn (approved):
requested some comments and one typo fix. otherwise good to go. Thank you!!

mod/wikirate/spec/lib/javascript/script_wikirate_common.test.coffee
xithan: sure, I'm not aware of any other options. I would prefer to use snake case for consistency with Ruby but since the rest of the world seems to use camel case in JS we don't really have a choice.

.pryrc
xithan: the file name is `pryrc` and whenever you enter a breakpoint you get a short description of all commands (the puts command at the bottom of this file) but of course an extra comment does't hurt.

.pryrc
```diff
@@ -0,0 +1,134 @@
...
+# Hit Enter to repeat last command
+Pry::Commands.command /^$/, "repeat last command" do
+  _pry_.run_command Pry.history.to_a.last
+end
+
+
+def hputs text
+  text = Nokogiri::XML(text, &:noblanks).root.to_s
+  puts CodeRay.scan(text, :html).term
+end
```
xithan: I totally forgot that I wrote that one. Good to know :smile: In situations where syntax highlighting would have been useful I couldn't remember the details of this nokogiri/coderay combo and I didn't want to spend the time to look it up.

ethn: yeah, very cool.  I just learned that teenagers are using "v" to mean very in conversation.

So I can just say "v kühl" to mean very cool.  Tip of the hat to Veronika.

mod/wikirate/spec/lib/javascript/script_wikirate_common.test.coffee
ethn: I think we should start shifting to camel in JS.  I kind of like that it helps signal a difference between JS and ruby.

.pryrc
ethn: I was thinking to make our breakpoint patterns (ie, the simple fact that pry is for breakpoints) more discoverable for new monkeys/platypi.

mod/wikirate/spec/lib/javascript/script_wikirate_common.test.coffee
xithan: "start shifting" to camel? In my perception we have always used camel in JS and only a few snake variables sneaked in because we don't have a hound for enforcing that rule.

ethn: fair enough. We just have to get over the camel's hump.

PR #866, opened 2018-07-19
adaption to decko's changed event api

xithan:
`skip: :allowed` instead of `optional: true`


ethn:
checked_by.rb is another natural candidate for this API. (eg `add_checked_flag?`)

If we don't do it as part of this PR, let's add a story.

PR #867, opened 2018-07-20
various renaming-related fixes

ethn:


mod/filter/set/abstract/01_filter_formgroups.rb
```diff
@@ -4,7 +4,7 @@
   end
 
   view :filter_year_formgroup, cache: :never do
-    select_filter :year, "Year", "most recent"
```
ethn, line 7: This stuff was all getting ignored by the API. I found that confusing.

xithan (approved):

PR #868, opened 2018-07-25
[Finishes #158737160] [Finishes #158738103]

dcastroeyss:


ethn:
please make sure PR is up to date (merge in master).  and we do not want to delete tmp/.gitkeep.  That is needed so that the tmp directory is created on new installations.

dcastroeyss:
tmp / .gitkeep was eliminated when running be rake wikirate: test: update_machine_output

.pryrc
```diff
@@ -1,168 +0,0 @@
-# These commands are available in the console when using binding.pry for breakpoints.
```
ethn: why delete this?

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -49,38 +72,62 @@ decko.editorContentFunctionMap['.wikiRating-editor'] = ->
...
     total += num * multiplier
-    valid = false unless num > 0
-  total = parseInt(total) / multiplier
-  publishWeightTotal(tbody, hash, total)
-  valid && total == 100
+    valid = false if num < 0 || !isMaxDigit(val)
+  if valid 
+    total = total / multiplier
+    publishWeightTotal(tbody, hash, total)
+    valid && total > 99.90 and total <= 100.09
```
ethn: since we're already within `if valid`, this `valid &&` is unnecessary, right?

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -49,38 +72,62 @@ decko.editorContentFunctionMap['.wikiRating-editor'] = ->
...
 validateWikiRating = (table) ->
   hash = wikiRatingEditorHash table
   valid = tallyWeights table, hash
+  #toEqualize table 
   updateWikiRatingSubmitButton table.closest('form.card-form'), valid
 
 
 DIGITS_AFTER_DECIMAL = 2
 
 tallyWeights = (tbody, hash) ->
```
ethn: this function is a little long for me.  We don't yet have strict rules for CoffeeScript (CodeClimate enforces rules for Ruby), but I start getting uncomfortable when a function is more than 5 lines, and usually I prefer 3 or less.

I would suggest separating this into two functions:  inputsAreValid and totalIsValid.

ethn: ...and you should write them so that they can both be tested with unit tests.

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -19,23 +19,46 @@ decko.slotReady (slot) ->
...
-  table.find("tbody tr").each ->
+  variableMetricRows(table).each ->
     cols = $(this).find('td')
     if (key = $(cols[0]).data('key'))
       hash[key] = $(cols[1]).find('input').val()
-  hash
+  hash 
+
+# if all values are equals return "true"
+valueFormEqual = (table) ->
```
ethn: Good comment.

Let's name this "variableValuesAreEqual".  We want to start using "is", "are", "has" etc in the names of methods that return true/false.

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -19,23 +19,46 @@ decko.slotReady (slot) ->
...
+  variableMetricRows(table).each ->
     cols = $(this).find('td')
     if (key = $(cols[0]).data('key'))
       hash[key] = $(cols[1]).find('input').val()
-  hash
+  hash 
+
+# if all values are equals return "true"
+valueFormEqual = (table) -> 
+  console.log('table', table)
```
ethn: this is for debugging, right?  please remove.

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -19,23 +19,46 @@ decko.slotReady (slot) ->
...
+  aux = false 
+  values = []
```
ethn: I suggest a separate method for this.  Eg.  "allVariableValues"

If you make one function `allVariableValues` that returns values, and a second function `allValuesAreEqual`, then you should be able to write a unit test for that second function, because it won't need the DOM.

ethn (changes requested):
Good stuff, Daniel.  If you're able to address the changes and write the tests I requested, I will deploy this to the dev server so that we can experiment with it in action.

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -72,15 +116,17 @@ DIGITS_AFTER_DECIMAL = 2
 
 tallyWeights = (tbody, hash) ->
```
ethn, line 113: This function is still too long. Please break it into multiple functions.  You should be able to make at least 2 (maybe more) functions that can be tested in Jest.  You do this by sending values (rather than DOM references) as objects.

mod/metrics/set/type_plus_right/metric/formula/wiki_rating.rb
```diff
@@ -18,7 +18,23 @@
   # table with Metrics on left and Weight inputs on right
   def rating_editor_table
     table rating_editor_table_content, class: "wikiRating-editor",
-                                       header: %w[Metric Weight]
+                                       header: ["Metric", checkbox_equalizer]
+  end
+
+  def checkbox_equalizer
```
ethn: perhaps "weight_heading" would be a clearer name? (with a comment that it includes the equalizer checkbox?

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -19,23 +19,45 @@ decko.slotReady (slot) ->
...
-  table.find("tbody tr").each ->
+  variableMetricRows(table).each ->
     cols = $(this).find('td')
     if (key = $(cols[0]).data('key'))
       hash[key] = $(cols[1]).find('input').val()
-  hash
+  hash 
+
+# if all values are equals return "true"
+variableValuesAreEqual = (table) ->
```
ethn: this is another case where you could make it testable if your argument were a list of values instead of a table.

We really want to simplify functions and add more tests so that the problem space is smaller if/when things break.

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -19,23 +19,46 @@ decko.slotReady (slot) ->
...
-  hash
+  hash 
+
+getValuesFromTable = (table) -> 
+  values = []
+  variableMetricRows(table).each -> 
+    tr = $(this) 
+    values.push( tr.find('td.metric-weight').find('input').val() )
+  values = values.splice(0, values.length - 1);
+  return variableValuesAreEqual(values)
```
ethn: Based on its name, this should return a list of values.  (That might be reusable).

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -49,18 +72,40 @@ decko.editorContentFunctionMap['.wikiRating-editor'] = ->
...
+  variableMetricRows(table).each ->
     tr = $(this)
     if key = tr.find(".metric-label .thumbnail").data "cardName"
       hash[key] = tr.find(".metric-weight input").val()
-  hash
+  hash 
+
+# if all values are equals active the equalize
+activeEqualize = () -> 
+  if !getValuesFromTable( $('.wikiRating-editor') )
```
ethn: I would suggest something like this:
```
values = getValuesFromTable( $('.wikiRating-editor') )
$('#equalizer').prop 'checked', variableValuesAreEqual(values)
```
... why do we call toEqualize if the values are already equal?

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -49,18 +72,40 @@ decko.editorContentFunctionMap['.wikiRating-editor'] = ->
...
+  else 
+    $('#equalizer').prop('checked', true)
+    toEqualize( $('.wikiRating-editor') )
+
+toEqualize = (table) -> 
+  val = (100 / (variableMetricRows(table).length - 1)).toFixed(2)
+    
+  variableMetricRows(table).each ->
+    tr = $(this)
+    tr.find('td.metric-weight').find('input').val(val)
```
ethn: this also looks like it should be a separate method to me.  like setAllVariableValuesTo(table, val)

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -72,15 +117,21 @@ DIGITS_AFTER_DECIMAL = 2
 
 tallyWeights = (tbody, hash) ->
   multiplier = 10**DIGITS_AFTER_DECIMAL
-  total =  0
+  aux = valuesAreValid(hash, multiplier)
+  if aux.valid 
+    total = aux.total / multiplier
+    publishWeightTotal(tbody, hash, total)
+    aux.valid = total > 99.90 and total <= 100.09
+  return aux.valid
```
ethn: I think this can be more compact, eg:
```
aux = valuesAreValid(hash, multiplier)
return false unless aux.valid
total = aux.total / multiplier
publishWeightTotal(tbody, hash, total)
total > 99.90 and total <= 100.09
```

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -72,15 +117,21 @@ DIGITS_AFTER_DECIMAL = 2
...
+
+valuesAreValid = (hash, multiplier) ->
```
ethn: This method should have a Jest test

mod/metrics/lib/javascript/script_metrics.js.coffee
```diff
@@ -19,23 +19,46 @@ decko.slotReady (slot) ->
...
+getValuesFromTable = (table) -> 
+  values = []
+  variableMetricRows(table).each -> 
+    tr = $(this) 
+    values.push( tr.find('td.metric-weight').find('input').val() )
+  values = values.splice(0, values.length - 1);
+  return variableValuesAreEqual(values)
+
+ # if all values are equals return "true"
+ variableValuesAreEqual = (values) ->
```
ethn: I think this can be shorter, just:
```
values.every( (val, i, arr) => val == arr[0] ) == true
```
no?  why is the aux needed/helpful?

I would like to see Jest tests for this.

mod/research/lib/javascript/script_research.js.coffee
```diff
@@ -18,6 +18,16 @@ decko.slotReady (slot) ->
 $(document).ready ->
   $("#main:has(>#Research_Page.slot_machine-view)").addClass("pl-0 pr-0")
 
+  $('#card_subcards__values_content').on "keyup", () ->  
+    if $(this).val().toLowerCase() != 'unknown'
+      $('#card_subcards__values_subcards__Unknown_content').prop('checked', false);
+    else 
+      $('#card_subcards__values_subcards__Unknown_content').prop('checked', true);
```
ethn: again could be tighter, eg:
```
selector = '#card_subcards__values_subcards__Unknown_content'
checked = $(this).val().toLowerCase() != 'unknown'
$(selector).prop 'checked', checked
```

ethn (changes requested):
almost there.  I'd really like to have some Jest tests.

mod/metrics/spec/lib/formula/javascript/script_metrics.test.coffee
```diff
@@ -0,0 +1,14 @@
+#require('../../../../lib/javascript/script_metrics.js.coffee')
+
```
xithan: I fixed in this PR #871 the issue why this test file didn't run.
(the error about "pairs-editor" : https://semaphoreci.com/ethan/wikirate/branches/pull-request-868/builds/21) 

But I don't think it's possible to access the methods that Ethan wants to be tested in a test file (like `valuesAreValid`). As far as I know you can call them only in the file where they are defined.
I guess we have to add them to `wikirate`, `decko` or `$` if we want to test them but it doesn't seem to be good practice to me to move these helper methods, that we use nowhere else, to those objects.

xithan: We could solve the problem with exporting the methods we want to test.
e.g. script_metric.js.coffee:
```coffee
exports.DIGITS_AFTER_DECIMAL = 2
exports.valuesAreValid = (hash, multiplier) ->
   ...
```
and then in script_metrics.test.coffee
`$.extend global, require('../../../../lib/javascript/script_metrics.js.coffee')`
```

ethn (approved):

PR #869, opened 2018-07-28
Vasso API needs

ethn:


mod/metrics/spec/set/type/metric/events_spec.rb
```diff
@@ -18,4 +18,21 @@
       end
     end
   end
+
+  describe "#delete answers" do
+    let(:metric) { Card["Joe User+researched number 1"] }
+    let(:delete_metric) { metric.update_attributes trigger: :delete_answers }
```
xithan: a method makes more sense if the return value is not used in the tests object

xithan (approved):

PR #874, opened 2018-08-15
things for number animation (so that vignesh verify this)

dcastroeyss:


mod/homepage/set/self/homepage_countup.rb
```diff
@@ -0,0 +1,10 @@
+include_set Abstract::CodeFile
```
srivig: this should be script_countup.rb , I am fixing this

mod/homepage/template/self/homepage_numbers.haml
```diff
@@ -4,10 +4,10 @@
   .mx-auto.mt-4
     .row
       .col-12.col-md-6
-        - categories.each do |cat|
+        - categories.each_with_index do |cat,i|
           .d-flex.mb-5.mr-4.justify-content-end
             .text-right.mx-3
-              %h1.font-weight-normal
+              %h1.font-weight-normal_count-ele{id: "count-ele-#{i}"}
```
srivig: missing `.` near `_count-ele`
Should be `%h1.font-weight-normal._count-ele{id: "count-ele-#{i}"}`

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -107,8 +118,32 @@ isScrolledIntoView = (elem) ->
   elemBottom = elemTop + $(elem).height();
   ((elemBottom <= docViewBottom) && (elemTop >= docViewTop));
 
-animation = (elem) ->
-  # animation CountUp.js
+runAnimation = (elem) ->
+  options = { useEasing: true, useGrouping: true, separator: ',', decimal: '.', };
+  demo = new CountUp($(elem).attr('id'),  0, parseInt($(elem).text()), 0, 3.5, options);
```
srivig: why `demo` ?

srivig: we need to remove the delimiter `number_with_delimiter(number, delimiter: ",")` from `hompeage_numbers.rb' ` in order make this countUp js work properly.

mod/homepage/template/self/homepage_numbers.haml
```diff
@@ -0,0 +1,21 @@
...
+          .d-flex.mb-5.mr-4.justify-content-end
+            .text-right.mx-3
+              %h1.font-weight-normal_count-ele{id: "count-ele-#{i}"}
+                = cat.count
+              %h5.font-weight-normal.text-muted
+                = cat.title
+            .drop-shadow.rounded-circle.bg-white.image-box.small.m-2{class: "#{cat.color}-color"}
+              = icon_tag(cat.icon, class: "md-24 p-3")
+      .d-none.d-md-inline.col-5.mt-4
+        .carousel.carousel-fade.slide.drop-shadow{"data-ride" => "carousel"}
```
srivig: missing `.` near `_count-ele`
Should be `%h1.font-weight-normal._count-ele{id: "count-ele-#{i}"}`

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -64,10 +107,48 @@ activateIntroTab = (tab)->
...
+isScrolledIntoView = (elem) ->
+  docViewTop = $(window).scrollTop();
+  docViewBottom = docViewTop + $(window).height();
+  elemTop = $(elem).offset().top;
+  elemBottom = elemTop + $(elem).height();
+  ((elemBottom <= docViewBottom) && (elemTop >= docViewTop));
+
+runAnimation = (elem) ->
+  options = { useEasing: true, useGrouping: true, separator: ',', decimal: '.', };
+  demo = new CountUp($(elem).attr('id'),  0, parseInt($(elem).text()), 0, 3.5, options);
```
srivig: we need to remove the delimiter `number_with_delimiter(number, delimiter: ",")` from `hompeage_numbers.rb' ` in order make this `parseInt($(elem).text())` work properly. number is already rendered with comma, so this parseInt won't work well. 
eg. 
output of parseInt("3434,5454")  will be 3434

srivig: also why `demo` ?

srivig (changes requested):

db/migrate_cards/20180815220000_new_homepage_codenames.rb
```diff
@@ -0,0 +1,41 @@
+# -*- encoding : utf-8 -*-
```
ethn: is this a new migration (and not a rename of the old?)

mod/homepage/set/self/script_countup.rb
```diff
@@ -0,0 +1,10 @@
+include_set Abstract::CodeFile
+Self::ScriptLibraries.add_item :script_countup
+
+def source_files
+  "vendor/CountUp.js/dist/countUp.js"
```
ethn: `CountUp.js` is not an acceptable name for a directory.  our directories are lower case and do not have extensions.

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -135,8 +149,32 @@ isScrolledIntoView = (elem) ->
...
+
+controlAnimate = (elem) ->
+  numberElementsControls.push( {id: $(elem).attr('id'), animated: false} )
+
+# has this element been animated?
+isAnimate = (id) ->
+  aux = false
+  numberElementsControls.forEach (elem) ->
+    if elem.id == id && elem.animated 
+      aux = true
```
ethn: does this actually return true?  It looks like it sets aux to true and then returns nothing.

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -135,8 +149,32 @@ isScrolledIntoView = (elem) ->
...
+runAnimation = (elem) ->
+  options = { useEasing: true, useGrouping: true, separator: ',', decimal: '.', };
+  animationNumber = new CountUp($(elem).attr('id'),  0, parseInt($(elem).text()), 0, 3.5, options);
+  animationNumber.start()
+
+controlAnimate = (elem) ->
+  numberElementsControls.push( {id: $(elem).attr('id'), animated: false} )
+
+# has this element been animated?
+isAnimate = (id) ->
```
ethn: isAnimate**d**

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -122,10 +135,11 @@ activateIntroTab = (tab)->
   active_panel.find('.carousel').carousel()
   active_panel.find('.carousel-item').first().addClass 'active'
 
-getNumbers = () -> 
+getNumberElements = () -> 
   values = []
-  $('.text-right.mx-3').each -> 
-    values.push( $(this).find('h1.font-weight-normal') )
+  $('h1.font-weight-normal._count-ele').each ->
```
ethn: `h1.font-weight-normal` is not a good selector.  It means the js will break if the style changes.

For javascript elements we typically prefer to use selectors that start with underscore. (and those selectors should NOT be used for styling)

mod/homepage/set/self/homepage_numbers.rb
```diff
@@ -22,6 +22,5 @@ def featured_answers
 
   def count card_type
     number = nest card_type, view: :count
```
srivig: remove `number` as it is not used anymore

srivig (changes requested):

mod/homepage/set/self/homepage_numbers.rb
dcastroeyss: `def count card_type
    nest card_type, view: :count
 end`

Yes?

ethn (changes requested):

db/migrate_cards/20180815220000_new_homepage_codenames.rb
dcastroeyss: It's a name change

mod/homepage/set/self/script_countup.rb
dcastroeyss: when added the git submodule was added with that name (CountUp.js)

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
dcastroeyss: ok

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
dcastroeyss: when "aux" is true then "return" stops the "forEach" and then the value of "aux" is returned

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
dcastroeyss: `$('h1._count-ele')`
yes?

db/migrate_cards/20180815220000_new_homepage_codenames.rb
ethn: Git doesn't think so.   Did you not delete the old file?

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
ethn: Why not just `_count-ele`? Is `_count-ele` used anywhere that isn't supposed to be counted?

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
ethn: ah, ok.  
This is so much cleaner in ruby ;)

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -135,8 +149,32 @@ isScrolledIntoView = (elem) ->
...
+isAnimate = (id) ->
+  aux = false
+  numberElementsControls.forEach (elem) ->
+    if elem.id == id && elem.animated 
+      aux = true 
+      return
+  aux  
+
+# this element is already animated now
+nowIsAnimate = (id) ->
```
ethn: wait.  what's the difference between this and isAnimated?

maybe the other should be `isAnimatable` and this is `isAnimated`?

sorry, I was confused before.

mod/homepage/set/self/script_countup.rb
ethn: I know.  please fix it.

db/migrate_cards/20180815220000_new_homepage_codenames.rb
dcastroeyss: No, I change the name

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
dcastroeyss: this function only sets the element as animated

dcastroeyss: isAnimated is to know if an element has already been animated

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
dcastroeyss: ok

mod/homepage/set/self/script_countup.rb
dcastroeyss: ok

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
ethn: oh, now that I'm looking at the functionality, I see that it's not a boolean call.  (And I should note that even if it returned true/false, usually a boolean call should not have side effects)

this should be something like `animateElem`, no

mod/homepage/lib/javascript/script_homepage_carousel.js.coffee
```diff
@@ -135,8 +149,32 @@ isScrolledIntoView = (elem) ->
...
+# has this element been animated?
+isAnimated = (id) ->
+  aux = false
+  numberElementsControls.forEach (elem) ->
+    if elem.id == id && elem.animated 
+      aux = true 
+      return
+  aux  
+
+# this element is already animated now
```
ethn: # animate this element

mod/homepage/vendor/CountUp/.editorconfig
```diff
@@ -0,0 +1,12 @@
+# This file is for unifying the coding style for different editors and IDEs
```
ethn: now this is not a submodule.

You need to make it a submodule with the correct directory name.  you can control the directory name. read about the `git submodule` command.

.gitmodules
```diff
@@ -17,6 +17,6 @@
 [submodule "mod/homepage/vendor/wodry"]
 	path = mod/homepage/vendor/wodry
 	url = https://github.com/srivig/wodry.git
-[submodule "mod/homepage/vendor/CountUp.js"]
-	path = mod/homepage/vendor/CountUp.js
-	url = https://github.com/inorganik/CountUp.js
+[submodule "mod/homepage/vendor/CountUp"]
+	path = mod/homepage/vendor/CountUp
+	url = https://github.com/inorganik/countUp.js.git
```
ethn: we control the name of the submodule and the path (and we want them to be lowercase).  we _don't_ control their url.

PR #877, opened 2018-08-22
Add "featured projects" to browse projects page

dcastroeyss:


mod/projects/set/self/project/core.haml
```diff
@@ -0,0 +1,14 @@
+%div
+  = field_nest :description
+  = nest :homepage_featured_projects, items: { view: "listing" }, wql: { limit: 3 }
```
ethn: we don't need the wql limit here.  We can show all of them.

mod/projects/set/self/project/core.haml
```diff
@@ -0,0 +1,14 @@
...
+%div{class: "tabbable", id: "project-tabs"}
+  %ul{class: "nav nav-tabs"}
+    %li.nav-item
+      %a{class: "active nav-link", href: "#project-active", data: {toggle: "tab"}} Active
+    %li.nav-item
+      %a{class: "nav-link", href: "#project-inactive", data: {toggle: "tab"}} Inactive
+  %div{class: "tab-content"}
```
ethn: You can actually shorten this line to just `.tab-content`

mod/projects/set/self/project/core.haml
```diff
@@ -0,0 +1,14 @@
+%div
+  = field_nest :description
+  = nest :homepage_featured_projects, items: { view: "listing" }
+%div{class: "tabbable", id: "project-tabs"}
```
ethn: again you don't need this %div.

Just `#project-tabs.tabbable` will do.  Please change all divs with classes

mod/projects/spec/set/self/project_spec.rb
```diff
@@ -0,0 +1,8 @@
+describe Card::Set::Self::Project do
+  specify "view core" do
+    expect_view(:core).to have_tag("div.card") do
```
ethn: this is confusing.  Why does core view have any of these classes?  It shouldn't have any of them.  Is this test actually passing??

mod/metrics/lib/javascript/script_value_type.js.coffee
```diff
@@ -11,7 +11,7 @@ showAndHide = (slot, value) ->
       when 'Number'
         'number_details'
       when 'Money'
-        'currency_details'
+        'number_details'
```
ethn: I can't approve this PR with this change in it, because it will break the value type functionality.

ALSO, this may be what is causing cukes to fail; you have to update the assets if you change js!

mod/metrics/set/type/metric/properties.rb
```diff
@@ -12,7 +12,9 @@ def research_properties
     {
       research_policy: "Research Policy",
       report_type:     "Report Type",
-      value_type:      "Value Type"
+      value_type:      "Value Type",
+      unit:            "Unit",
+      range:           "Range"
```
ethn: same here and in display.rb.  we can't approve half completed functionality

ethn (changes requested):

mod/projects/set/self/project/core.haml
```diff
@@ -0,0 +1,14 @@
...
+  = field_nest :description
+  = nest :homepage_featured_projects, items: { view: "listing" }
+#project-tabs.tabbable
+  %ul.nav.nav-tabs
+    %li.nav-item
+      %a{class: "active nav-link", href: "#project-active", data: {toggle: "tab"}} Active
+    %li.nav-item
+      %a{class: "nav-link", href: "#project-inactive", data: {toggle: "tab"}} Inactive
+  .tab-content
+    #project-active.tab-pane.active
```
ethn: as I've said a couple of times, this should loop through all options cards.

mod/projects/spec/set/self/project_spec.rb
```diff
@@ -0,0 +1,8 @@
+describe Card::Set::Self::Project do
+  specify "view core" do
+    expect_view(:core).to have_tag("div.card-slot") do
```
ethn, line 3: I don't understand why this passes.  Why does the outer div have a card-slot?

ethn (changes requested):

mod/projects/set/self/project/core.haml
ethn: Here's what I want you to do.

Warning: some of this is fun ;)

1. using staging or dev, do some sharking, and make it so that any `[Status]+Projects` card will produce a search for projects with that status.  For example, `Active+Projects` will give you the same thing that `Project+Active` currently does.  You do this by creating a `*structure` rule that applies to `[Status]+Projects`.  You will need to use WQL syntax.  We use that all the time, so this will be good practice.
2. Once you have that working, I want you to add all the cards you used to make this work to the "export" card.
3. Then you can use `be rake wikirate:import_from_local  name=import_new_project_status` to pull down the contents of the export card to a migration. This will make it so that when we deploy live, it will add all those cards.  (You can use anything for the name as long as it starts with `import_`)
4. Then, in the ruby code, you will need to get a list of all the status options for projects.  See `https://wikirate.org/Project+status+*type_plus_right+*options`.  I think I've already added a comment somewhere discussing this more.
5. Then in the HAML, you will need to loop through each of those options, and then nest the `+:project` card in it.

If you do this, then all someone has to do is add a new Status to the options list (which will happen soon), and then the interface will handle it all without any additional coding.

mod/projects/spec/set/self/project_spec.rb
ethn: OK, now I understand how this passes.

It looked to me like this was testing to see whether there was a .card-slot with a .tabbable and a .tab-content inside it.  There isn't, but that's not what this is testing.   To test that, the 2nd and 3rd `have_tag` would need to be `with_tag`.

As written, the test is just testing whether each of those three divs exist at all, and they do (just not inside each other)

If you just want to test that they exist, try:
```
  expect_view(:core).to have_tag("div.card-slot") 
                    .and have_tag("div.tabbable")
                    .and have_tag("div.tab-content")
```

... or, if you want to look for a div that is outside all three of them, you need to add a class to the outer div in the haml.

In either case, I think it would be wise to make sure that the test confirms that we are seeing the featured projects card.  (and I assume you've confirmed in the browser that this all looks right)

mod/metrics/set/type/metric/display.rb
```diff
@@ -57,8 +57,7 @@ def category_legend
     voo.hide :menu
     wrap do
       [
-        field_nest(:value_type, view: :content, items: { view: :name }, show: :menu),
-        _render_short_view
+        field_nest(:value_type, view: :content, items: { view: :name }, show: :menu)
```
ethn: this is a value type change (doesn't belong here)

ethn (changes requested):
I added some requested changes, but I think we could in principle deploy this as is (if you removed the value type change).

Would you prefer that we merge this and then do a new pull request with the additional changes or wait until it's all done?

dcastroeyss:
I will make a new PR with the required changes

ethn:
closing (these changes updated unified in another branch)

PR #879, opened 2018-08-24
Type value

dcastroeyss:
simplify value type handling


mod/metrics/set/type/metric/properties.rb
```diff
@@ -45,6 +47,14 @@ def metric_type_property
     field_nest :metric_type, view: :content, items: { view: :name }
   end
 
+  def unit_property
+    field_nest :unit, view: :content, items: { view: :name }
+  end
+
+  def range_property
+    field_nest :range, view: :content, items: { view: :name }
+  end
```
ethn, line 57: this looks like good progress.  of course we need the new javascript to handle showing/hiding.

dcastroeyss: that is not handled in `script_value_type.js.coffee` Lines 8/18?

ethn: If you've tried it and it works, then I'll have a look.

Normally we don't accept code changes without a test that proves that they work.  It would be good to have a cucumber feature for this.

mod/metrics/set/type/metric/properties.rb
```diff
@@ -12,7 +12,9 @@ def research_properties
     {
       research_policy: "Research Policy",
       report_type:     "Report Type",
-      value_type:      "Value Type"
+      value_type:      "Value Type",
+      unit:            "Unit",
+      range:           "Range"
```
ethn: what about options?

dcastroeyss: options_property?

mod/metrics/lib/javascript/script_value_type.js.coffee
```diff
@@ -5,18 +5,44 @@ showField = (divName) ->
...
+  showOrHideProperty(properties, 'show') if properties.length > 0
+
+showOrHideProperty = (properties, option) ->
+  element = null
+  properties.forEach (value) ->
+    selector = '#cdp-scope_2_emission-detail_tab td:contains('+value+')'
+    if $($(selector)[1]).parent().length > 0
+      element = $($(selector)[1]).parent()
+    else 
+      element = $($(selector)[0]).parent()
```
ethn: Without looking at the HTML, I really don't have any idea what this does.  We need some better variable names that explain what's happening.

mod/metrics/lib/javascript/script_value_type.js.coffee
```diff
@@ -5,18 +5,44 @@ showField = (divName) ->
...
+    if option == 'show' then element.show() else element.hide()
+
 showAndHide = (slot, value) ->
   div_to_show =
     switch value
-      when 'Number'
+      when 'Number', 'Money'
         'number_details'
-      when 'Money'
```
ethn, line 13: note to @xithan - we're changing the logic here so that money values have a `unit` card rather than a `currency` card.  we're going to need to address this in several places (and include a migration) before we deploy, but I wasn't sure whether it made sense to assign this to Daniel given how distributed that work is.

ethn (changes requested):
showOrHideProperty needs improved readability.

mod/metrics/lib/javascript/script_value_type.js.coffee
```diff
@@ -5,18 +5,44 @@ showField = (divName) ->
...
+showAndHideProsTable = (value) ->
+  properties = ['Unit','Range','Options']
+  showOrHideProperty(properties, 'hide')
+  properties = getPropertiesValueType(value)
+  showOrHideProperty(properties, 'show') if properties.length > 0
+
+showOrHideProperty = (properties, option) ->
+  element = null
+  properties.forEach (value) ->
+    selector = '#cdp-scope_2_emission-detail_tab td:contains('+value+')'
```
dcastroeyss: I do not like this selector, I used it because I did not find another

mod/metrics/lib/javascript/script_value_type.js.coffee
dcastroeyss: element refers to the html element that will be shown or hidden

mod/metrics/lib/javascript/script_value_type.js.coffee
ethn: you can add a better one.

mod/metrics/lib/javascript/script_value_type.js.coffee
ethn: well obviously.  but I have no idea what the HTML looks like, and it looks very fragile. someone could change the HTML and it would break everything.

mod/metrics/lib/javascript/script_value_type.js.coffee
dcastroeyss: with the HTML structure I have not found a better one

mod/metrics/lib/javascript/script_value_type.js.coffee
dcastroeyss: I will try to do something better. But I think that the possibility of changing the HTML and the selectors stop working is always present

ethn: That's why you write tests Daniel.  Write a cucumber test that proves this works.  Then, if somebody changes the HTML and breaks it, the cucumber test will tell us.

mod/metrics/lib/javascript/script_value_type.js.coffee
ethn: Daniel, you can add any selectors you want.  In fact, we want javascript to be using (almost) exclusively selectors that start with "_".  So you need to add some here.

You do understand that `#cdp-scope_2_emission-detail_tab` will only show up on the metric named "cdp scope 2 emissions", right?  Your code will work on, at best, ONE metric.

But even if you didn't add selectors, you'd need to write better code.  Instead of saying `$selector[1]` three times, assign it to a variable name that tells us what it is.  These are very abstract terms.  I should be able to understand your code without having to look at the HTML.

The same goes for this: 

```
    if $($(selector)[1]).parent().length > 0
      element = $($(selector)[1]).parent()
```

a good variable would make your code more readable, more maintainable, and (ever so slightly) faster.

ethn (changes requested):

mod/metrics/lib/javascript/script_value_type.js.coffee
dcastroeyss: in what file should I write the cucumber test for this?

ethn: somewhere in mod/metrics/features.  It's probably worth its own feature: `metric_edit.feature`

mod/projects/set/self/project/core.haml
```diff
@@ -0,0 +1,14 @@
...
+  = nest :homepage_featured_projects, items: { view: "listing" }
+#project-tabs.tabbable
+  %ul.nav.nav-tabs
+    %li.nav-item
+      %a{class: "active nav-link", href: "#project-active", data: {toggle: "tab"}} Active
+    %li.nav-item
+      %a{class: "nav-link", href: "#project-inactive", data: {toggle: "tab"}} Inactive
+  .tab-content
+    #project-active.tab-pane.active
+      = nest "Project+Active", title: "Active Projects", items: { view: "listing" }
```
ethn: this still has Active and Inactive hard coded.  We need to:
1. get the options from the options rule in Project+status
2. create +Project cards on each status with a +*type_plus_right rule.  Eg, Active+Projects should give you a list of all the active projects
3. make this file loop through all the statuses.

ethn:
closing (these changes updated unified in another branch)

PR #881, opened 2018-08-28
use new card-mod "csv_import"

xithan:


mod/wikirate_csv_import/set/type_plus_right/source/file/import_answers_from_source.rb
```diff
@@ -0,0 +1,77 @@
+include_set Abstract::ImportWithCompanies
```
ethn, line 1: is this new?

ethn (approved):

mod/wikirate_csv_import/set/type_plus_right/source/file/import_answers_from_source.rb
xithan: it's old but could be that you deleted it during your source clean up

PR #883, opened 2018-08-31
value cardtypes

ethn:


mod/metrics/set/type/category_value.rb
```diff
@@ -0,0 +1,21 @@
+include_set Abstract::Pointer
```
ethn, line 2: only the pointer-like value types have pointer methods now.

mod/wikirate_source/set/source_type/file.rb
```diff
@@ -8,10 +8,4 @@
   def icon
     "upload"
   end
-
-  view :metric_import_link do
```
ethn, line 12: not really related. see https://www.pivotaltracker.com/story/show/160198888

xithan (approved):
I reviewed this in the other PR

PR #884, opened 2018-09-03
[Finishes #159936842]

srivig:
Fixed safari bug


ethn (approved):
I wouldn't want a lot of browser-specific code throughout the site, but I think we can make an exception for the homepage.

@xithan, are you happy with this?

xithan:
I prefer a more declarative PR title. Afaik the PT code in the PR title doesn't have any effect on the automatic integration. What would have a big impact on the integration is if we were adding the PT code to the branch name like this: 123456-branch-name. (see https://www.pivotaltracker.com/integrations/GitHub) I only did that once my self to test it and it worked. I think we should try to make that a habbit.

xithan (approved):

PR #885, opened 2018-09-10
industries (so far) and error handling

ethn:


mod/metrics/set/abstract/value.rb
```diff
@@ -28,68 +32,10 @@ def company_id
...
-
-event :mark_as_imported, before: :finalize_action, when: :import_act? do
-  @current_action.comment = "imported"
-end
-
-# def answer_imported?
-#   ActManager.act_card.type_id.in? [AnswerImportFileID, RelationshipAnswerImportFileID]
+# def record
+#   name.parts[0..-3].join "+"
+# end
```
xithan: seems like good api to me. Why is it commented?

mod/metrics/set/type/free_text_value.rb
```diff
@@ -0,0 +1,11 @@
+include_set Abstract::Value
+
+format :html do
+  view :editor do
+    text_field :content, class: "d0-card-content"
+  end
+
+  def pretty_value
+    @pretty_value ||= card.value
+  end
```
xithan: that could be the default in `Abstract::Value`

mod/metrics/set/type/metric/creation.rb
```diff
@@ -78,7 +78,7 @@ def create_value_args args
     type_id: type_id,
     "+value" => {
       content: args[:value],
-      type_id: (args[:value].is_a?(Integer) ? NumberID : PhraseID)
+      type_code: value_cardtype_code
```
xithan: dot missing ?!

mod/isic/spec/set/type/isic_class_value_spec.rb
```diff
@@ -0,0 +1,5 @@
+# -*- encoding : utf-8 -*-
+
+RSpec.describe Card::Set::Type::IsicClassValue do
+
```
xithan: would be good to have at least one test that checks that pulling the options from json is working

mod/metrics/set/type/metric/answers.rb
```diff
@@ -70,16 +71,23 @@ def value_type_code
   value_type_card.item_cards.first&.codename || default_value_type_code
 end
 
+def value_cardtype_code
+  :"#{value_type_code}_value"
+end
```
xithan: can we document somewhere what all these variants of compositions of "value", "code", and "type" mean and what kind of value they hold?

xithan (changes requested):
"If it's not tested, it doesn't work"

mod/metrics/set/type/metric/creation.rb
xithan: probably correct this way

mod/isic/spec/set/type/isic_class_value_spec.rb
ethn: agree. all that code is now in decko, though.  will write it there.

mod/metrics/set/abstract/value.rb
ethn: I commented because it wasnt used.  but I restored it.

ethn:
The JSON options are now entirely handled (and lightly tested) in Decko.

xithan (approved):

PR #893, opened 2018-09-13
refactor latest research page changes

xithan:


mod/research/set/self/research_page.rb
```diff
@@ -31,110 +31,4 @@ def slot_machine opts={}
       haml :slot_machine
     end
   end
```
xithan, line 33: no changes here. I only separated code for the left and the right side of the research page

mod/wikirate/set/all/wikirate.rb
```diff
@@ -16,6 +16,8 @@
     voo.show :horizontal_menu if main?
   end
 
+  NEW_BADGE = '<span class="badge badge-danger">New</span>'.freeze
```
ethn, line 19: just here for safekeeping?

mod/research/set/self/research_page/navigation.rb
```diff
@@ -6,14 +6,18 @@ def next_button type
...
-                  class: "btn btn-outline-secondary btn-xs-icon mx-2"
+  def navigate_button type, text, item, extra_class=nil
+    opts = { class: "btn btn-outline-secondary btn-xs-icon" }
+    add_class opts, extra_class if extra_class
+    if item
+      opts[:path] = research_url(type => item)
+    else
+      add_class opts, "disabled"
+    end
+    link_to text, opts
```
ethn, line 20: I take it we decided to make the navigate button always there (but disabled when not applicable?)

ethn (approved):

mod/research/set/self/research_page/navigation.rb
xithan: that was the quickest fix to avoid that the buttons move around when one of them is missing

mod/wikirate/set/all/wikirate.rb
xithan: yes

PR #894, opened 2018-09-19
reorganize metrics mod

xithan:
I separated a `calculated_metrics` and a `researched_metrics` mod from the `metrics_mod`


mod/calculated_metrics/set/abstract/value/update_calculations.rb
```diff
@@ -0,0 +1,11 @@
...
+  ensure_metric(metric_card).each_dependent_score_metric do |metric|
+    metric.update_value_for! company: company_id, year: year
+  end
+end
+
+event :update_related_calculations, :after_integrate, skip: :allowed do
+  ensure_metric(metric_card).each_dependent_formula_metric do |metric|
+    metric.update_value_for! company: company_id, year: year
+  end
+end
```
ethn, line 11: what's the rationale for separating this from the other events?

ethn (approved):

mod/calculated_metrics/set/abstract/value/update_calculations.rb
xithan: it's about calculated metrics

PR #899, opened 2018-09-27
privacy issues and script cleanup

ethn:
I'm happy to restore any script here with code that you think will be useful as a reference to us in the future, but I wanted to clean up a bit.

We could also consider moving them to an "old" directory...


mod/wikirate/set/all/wikirate/google_analytics.rb
```diff
@@ -1,5 +1,5 @@
 format :html do
   def google_analytics_snippet_vars
-    super.insert 1, [:_setPageGroup, 1, card.type_name]
+    super.insert(1, [:_setPageGroup, 1, card.type_name]).insert 2, ['_gat._anonymizeIp']
```
ethn: IP anonymization

script/cron/delete_old_ip_addresses.rb
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
+
+require File.expand_path("../../../config/environment", __FILE__)
+
+Card::Act.where("acted_at < ?", 6.months.ago).update_all ip_address: nil
```
ethn: deleting old IP addresses.

script/cron/table_maintenance.rb
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby
 
-require File.dirname(__FILE__) + "/../config/environment"
+require File.expand_path("../../../config/environment", __FILE__)
```
ethn, line 3: note: will need to update cron jobs upon deploy!

VERSION
```diff
@@ -1 +1 @@
-0.15.0
```
ethn, line 1: wait, we never deployed 0.15.0, so this tick shouldn't happen yet.  will fix

xithan:
I don't see any value in removing all the scripts. I would keep them.

xithan (approved):

PR #900, opened 2018-09-28
cleanup of claims, analyses, etc.

ethn:


mod/metrics/set/self/script_metrics.rb
```diff
@@ -6,7 +6,6 @@
     metrics
     metric_properties
     metric_value
-    drag_and_drop
     metric_chart
```
xithan, line 9: Didn't you remove that file?

mod/metrics/lib/javascript/script_metric_chart.js.coffee
```diff
@@ -1,31 +0,0 @@
...
-  decko.filterCategorySelected($('a[data-category="metric_value"]'))
-
-metric_chart = (spec, id) ->
-  runtime = vega.parse spec
-  view = new vega.View(runtime).initialize($("##{id}")[0]).hover().run()
-  view.addEventListener('click', (event, item) ->
-    if item.datum.link
-      $.ajax item.datum.link,
-        success: (data) -> $(event.target).setSlotContent(data)
-  )
```
xithan: Are you sure that we don't need that file anymore?

xithan (changes requested):

mod/metrics/set/self/script_metrics.rb
ethn: not intentionally.  good catch.

xithan (approved):

PR #902, opened 2018-09-28
unify wolfram and ruby formulas

xithan:
[Finishes #158314840]


mod/calculated_metrics/lib/formula/wolfram.rb
```diff
@@ -1,23 +1,34 @@
+require_dependency "wolfram/unknowns"
+
```
ethn, line 2: nice refactoring!

ethn (approved):

PR #904, opened 2018-10-08
answer counts, etc

ethn:


mod/caching/set/type_plus_right/metric/source.rb
```diff
@@ -2,14 +2,20 @@
...
-  { referred_to_by: {
-    right: "source",
-    left: {           # answer
-      left: {         # record
-        left: "_left" # metric
-      }
-    }
-  } }
+  { referred_to_by: { right_id: Card::SourceID, left_id: [:in] + answer_ids } }
+end
```
ethn, line 6: optimization.

mod/caching/set/type_plus_right/wikirate_company/wikirate_topic.rb
```diff
@@ -16,64 +16,59 @@ def company_name
...
+  ::Answer.select(:metric_id).where(company_id: left.id).distinct.pluck(:metric_id)
   # pluck seems dumb here, but .all isn't working (returns *all card)
 end
 
-# @returns Array [[topic_card1, num_metrics], [topic_card2...]...] }
-def topics_by_metric_count
-  topic_metric_count_hash.sort_by { |_card, count| count }.reverse
-end
-
+# # @returns Array [[topic_card1, num_metrics], [topic_card2...]...] }
```
ethn, line 42: temporarily taking this off.  the topics/company connection may need special handling so it can show topic-specific counts on companies and vice versa.  but for now it's more important to move to the bar/box design, which won't really support all of this yet.

mod/metric_answer_lookup/lib/answer/cardless_answers.rb
```diff
@@ -42,8 +42,11 @@ def calculated_answer metric_card, company, year, value
     end
 
     def update_cached_counts
-      [[metric_id, :value], [metric_id, :wikirate_company],
-       [company_id, :metric], [company_id, :wikirate_topic]].each do |mark|
+      [[metric_id, :metric_answer],
+       [metric_id, :wikirate_company],
+       [company_id, :metric],
+       [company_id, :metric_answer],
+       [company_id, :wikirate_topic]].each do |mark|
```
ethn, line 49: would be nice if this were a list you could register to from the relevant set.

mod/metrics/lib/card/metric.rb
```diff
@@ -1,41 +1,60 @@
...
+      prep_source args
+      args
+    end
+
+    def prep_source args
+      return unless @metric.researchable? && @test_source
+      args[:source] ||= test_source_card
+    end
+
+    def test_source_card
```
ethn, line 48: most of the above is just clean up, but the impetus was to move from "random source" to a specific set of sources. without that, testing source counts wasn't reliable.

PR #905, opened 2018-10-10
emergency fixes to error handling and metric properties

ethn:


ethn:
I want to rush this through emergency deploy.  The value_options thing isn't a huge deal, but the error handling is creating a LOT of noisy errors on new relic.

PR #910, opened 2018-10-29
bar box tweaks and calculation fixes

ethn:


mod/calculated_metrics/set/abstract/value.rb
```diff
@@ -0,0 +1,5 @@
+event :update_related_calculations, :after_integrate, skip: :allowed do
```
ethn, line 1: need specs for this.

ethn: (done)

mod/metrics/set/type/metric/tabs.rb
```diff
@@ -19,7 +19,7 @@ def tab_list
   end
 
   def nest_about
-    return "" if card.about_card.new?
+    # return "" if card.about_card.new?
```
ethn, line 22: Laureen specifically requested the "add about" functionality.  not sure why we removed it...

PR #916, opened 2018-11-14
make relationship answers editable again

xithan:


mod/research/set/self/research_page/params.rb
```diff
@@ -183,7 +187,14 @@ def record_card
...
-    @sac ||= Card.fetch [metric, company, year.to_s], new: { type_id: MetricAnswerID }
+    @sac ||=
+      if related_company
+        Card.fetch [metric, company, year.to_s, related_company],
+                   new: { type_id: RelationshipAnswerID }
+      else
+        Card.fetch [metric, company, year.to_s], new: { type_id: MetricAnswerID }
+      end
+
   end
```
ethn, line 198: We don't have to do very sophisticated testing, but can we just make sure the edit view for relationship answers doesn't rat?

ethn (changes requested):
looks good. one spec request.

ethn (approved):

PR #921, opened 2018-12-01
pdf conversion 

ethn:


ethn:
Note: this includes Vignesh's PR.

initializers/pdfkit.rb
```diff
@@ -0,0 +1,9 @@
+require "bundler"
+
+PDFKit.configure do |config|
+  config.default_options = {
+    page_size: "A4",
+    print_media_type: false
+  }
+  config.wkhtmltopdf = "#{Bundler.bundle_path}/bin/wkhtmltopdf"
```
ethn, line 8: PDFKit wasn't finding wkhtmltopdf without this.  I don't think it's been working at all on the deployed sites.

lib/tasks/wikirate/test.rake
```diff
@@ -63,6 +63,7 @@ namespace :wikirate do
 
     desc "update caches for machine output"
     task update_machine_output: :environment do |task|
+      ENV["STORE_CODED_FILES"] = "true"
```
ethn, line 66: it may make more sense to do this as a Cardio.config option.  

The problem is that some sites (eg UBOS) were trying to overwrite gem files.

xithan (approved):

PR #923, opened 2018-12-05
metric import

xithan:


script/csv_import/metric/metric_csv_row.rb
```diff
@@ -34,19 +24,49 @@ def normalize_research_policy value
...
       new_value
     end
   end
 
+  def format_html html
+    html.gsub(/\b(OR|AND)\b/, "<strong>\\1</strong>")
+        .gsub(/Note:([^<]+)<br>/, "<em><strong>Note:</strong>\\1</em><br>")
+        .gsub(/<p>([^<]+)<br>/, "<p><strong>\\1</strong><br>")
+        .gsub("Sources:", "<strong>Sources:</strong>")
+        .gsub(/(<br><br>|^)([^<]+)(?=<br>)/) do |m|
```
ethn, line 45: might read a bit better as key value pairs.
```
HTML_SUBSTITUTIONS = {
  /\b(OR|AND)\b/ =>  "<strong>\\1</strong>"
...
}
```

ethn (approved):

script/csv_import/metric/metric_csv_row.rb
xithan: If the last one could be handled that way too, then I would be willing to change it.

PR #924, opened 2018-12-06
little import tweaks

xithan:


mod/wikirate_csv_import/set/self/import_tool.rb
```diff
@@ -21,13 +21,19 @@ def tab_content key
...
 
   def new_import_form import_type
-    nest(Card.new(type: import_type),
-         view: :new, hide: [:header, :menu, :new_type_formgroup])
+    if ok? :create
+      nest(Card.new(type: import_type),
+           view: :new, hide: [:header, :menu, :new_type_formgroup])
+    else
+      "You have to #{signin_link} to create new import files."
+    end
```
ethn, line 36: doesn't the :new view already handle this with perms: :create ?

ethn: ...oh, it prompts to add.  This is part of the unknown API I want to improve...

ethn (approved):

mod/wikirate_csv_import/set/self/import_tool.rb
xithan: no, it's just blank

PR #926, opened 2018-12-12
network aware metrics

xithan:
introduces also relationship answer lookup table and changes the whole formula parsing process
(both inevitable to make network aware metrics work)

TODO: 
- user documentation and code documentation for network aware metrics
- yearly variables don't work with a single company option (a very exceptional use case)


db/migrate_cards/20181212123127_add_always_one_yearly_variable.rb
```diff
@@ -0,0 +1,15 @@
+# -*- encoding : utf-8 -*-
+
+class AddAlwaysOneYearlyVariable < Card::Migration
+  def up
+    subcards =
+      Card.search(typy_id: Card::YearID, return: :name)
```
ethn: type/typy/typo

mod/caching/set/all/relationship_cache.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+end
+
+def act_based_refresh_of_relationship_lookup_entry ids
+  @updated_relationships ||= ::Set.new
+  @updated_relationships.merge ids
+end
+
+event :refresh_updated_relationships, :integrate, after: :refresh_updated_answers do
+  return unless @updated_relationships.present?
```
ethn: better in `:when`?

(general principle is that putting it there is better practice so it won't trigger other callbacks)

mod/calculated_metrics/chunk/formula_input.rb
```diff
@@ -3,6 +3,7 @@ class Content
     module Chunk
       class FormulaInput < Nest
         Card::View::Options.add_option :year, :ruler
```
ethn, line 5: we need to change "ruler" to "shark" :)

mod/calculated_metrics/lib/formula/calculator.rb
```diff
@@ -11,14 +11,17 @@ module Formula
...
-      @input = Input.new @formula_card.input_cards,
-                         @formula_card.input_requirement,
-                         year_options,
-                         &self.class::INPUT_CAST
+      @input =
+        if formula_card.input_cards.any?(&:nil?)
+          InvalidInput.new
+        else
+          ::Formula::Calculator::Input.new @formula_card, &self.class::INPUT_CAST
+        end
```
ethn: seems like this could be in a lazy loading `def input` method.

mod/calculated_metrics/lib/formula/calculator.rb
```diff
@@ -99,14 +102,6 @@ def compile_formula
...
     end
 
-    # Extracts all year options from all input nests in the formula
-    def year_options
-      @year_options ||=
-        @formula_card.input_chunks.map do |chunk|
-          chunk.options[:year]
-        end
-    end
-
```
ethn, line 109: so the comment at the top of the file about needing to respond to `input_chunks` is no longer valid, right?  now it needs to respond to `year_options`?

mod/calculated_metrics/lib/formula/calculator/companies_with_values.rb
```diff
@@ -0,0 +1,38 @@
+module Formula
+  class Calculator
+    # Keeps track of companies that have values for all input metrics which means
+    # it's possible to calculate answers for them.
+    # It's a Set of company ids for every year
+    class CompaniesWithValues < Hash
```
ethn, line 6: YES!  May the Lord of Platypuses bless you for doing this.

mod/calculated_metrics/lib/formula/calculator/input.rb
```diff
@@ -7,18 +7,15 @@ class Calculator
...
-      def initialize input_cards, requirement, year_options, &input_cast
-        @input_cards = input_cards
-        @requirement = requirement
+      def initialize formula_card, &input_cast
+        @input_cards = formula_card.input_cards
+        @requirement = formula_card.input_requirement
         @input_cast = input_cast
-        @year_options_processor = YearOptionsProcessor.new year_options
-        @input_values = initialize_input_values
+        @input_values = InputValues.new formula_card
```
ethn, line 18: seems like at least the first 2 (and maybe all but input_cast) could be done with lazy loading.  What I like about that is that you could add accessors and then you could write tests that tweak those values without having to mess with the formula card.

mod/calculated_metrics/lib/formula/calculator/input_item.rb
```diff
@@ -0,0 +1,130 @@
...
+    # {MetricInput} and {YearlyVariableInput}.
+    # The logic for the nest options is in the modules {}CompanyOption} and {YearOption}
+
+    # Note: The main difference between metrics and yearly variables is that
+    # yearly variables are in general independent of the company of the answer we
+    # want to calculate.
+    # But this can change. A metric with a fixed company option is also company
+    # independent where as a yearly variable with a related company option is company
+    # dependent. That's why the company dependency is separated into the modules
+    # {CompanyDependentInput} and {CompanyIndependentInput}
```
ethn, line 21: VERY helpful comments.  Thank you.

mod/calculated_metrics/lib/formula/calculator/input_list.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      def add_item i
+        item = item_class(i).new(@input_values, i, @input_values.all_input_required?)
+        self << item
+      end
+
+      def item_class i
+        input_card = @input_values.input_cards[i]
+        binding.pry if input_card.nil?
```
ethn: binding.pry

mod/calculated_metrics/lib/formula/calculator/input_item/year_option.rb
```diff
@@ -0,0 +1,72 @@
...
+          interpret_year_option
+          processed_year_option
+        end
+
+        def full_search
+          values_by_year_for_each_company do |c_id, values_by_year|
+            years = values_by_year.keys
+            translate_years(years).each do |year|
+              final_val = apply_year_option values_by_year, year
+              store_value c_id, year, final_val
```
ethn, line 30: my `unknown` solution involved an interjection "here".  My notion was not to store the value in the first place if the strategy was `:reject`.  And if it was `string` or `null` or whatever, then it seemed like a good idea to do the conversion before even storing it.  (my notion was that different calculators could then cast those values differently later on.  Eg wolfram would cast `nil` as `"null"`.

mod/calculated_metrics/lib/formula/calculator/input_item/value_store.rb
```diff
@@ -0,0 +1,30 @@
...
+          @values = Hash.new_nested(Hash, Hash)
+        end
+
+        def get company, year=nil
+          dig_args = [company, year].compact
+          dig_args.present? ? values.dig(*dig_args) : values
+        end
+
+        def companies
+          @values.keys
```
ethn: @companies is a Set, but `#companies` is an Array?

ethn: I'm not seeing what `@companies` does here...

mod/calculated_metrics/lib/formula/calculator/input_item/value_store.rb
```diff
@@ -0,0 +1,30 @@
...
+    class InputItem
+      # Used to hold onto the values that an input item finds
+      class ValueStore
+        attr_reader :years, :values
+        def initialize
+          @years = ::Set.new
+          @companies = ::Set.new
+          @values = Hash.new_nested(Hash, Hash)
+        end
+
```
ethn, line 12: can we add a `@return` comment here?

If I understand correctly, it returns:
`[Hash] { company_id => { year => value }` if company is nil and year is nil or missing
`{ year => value }` if company is valid and year is nil or missing,
`value` if both are present.

I think it would also try to look up company_id based on a year if company is nil and year is valid, but that's an edge case :)

mod/calculated_metrics/lib/formula/calculator/input_item/year_option.rb
ethn: hmm, maybe here's not quite right.  maybe we need a store_normalized_value method.  Does seem like normalization makes more sense in input_values than in input...

mod/calculated_metrics/lib/formula/calculator/input_item/yearly_value_store.rb
```diff
@@ -0,0 +1,22 @@
+module Formula
+  class Calculator
+    class InputItem
+      # Provides the same api as ValueStore but ignores the company arguments
+      class YearlyValueStore < ValueStore
+        def initialize
+          @years = ::Set.new
+          @values = Hash.new_nested(Hash)
+        end
+
```
ethn, line 11: @return comment here too?

mod/calculated_metrics/lib/formula/calculator/invalid_input.rb
```diff
@@ -0,0 +1,14 @@
+module Formula
+  class Calculator
+    # Placeholder for the case that formula contains input items that don't exist
```
ethn, line 3: I like this.  Probably good to have it before we recruit a lot more people to write formulae!

ethn (approved):
exquisite stuff.

mod/caching/set/all/relationship_cache.rb
xithan: ✅

mod/calculated_metrics/lib/formula/calculator.rb
xithan: right, and `company_options`

PR #931, opened 2018-12-20
network aware update

xithan:
improved validations
unknown and not_researched options
CountRelated function
a lot of refactoring


mod/calculated_metrics/lib/formula/calculator/input_item/options/not_researched_option/not_researched_pass_value.rb
```diff
@@ -0,0 +1,42 @@
...
+              end
+            end
+
+            def mandatory?
+              false
+            end
+
+            def sort_index
+              super + 2 * parser.input_count
+            end
```
ethn, line 36: maybe a comment here?

ethn (approved):

mod/calculated_metrics/lib/formula/calculator/input_item/options/not_researched_option/not_researched_pass_value.rb
xithan: sure (there is a comment where `sort` is called  which gives you at least a chance to understand what this is about)

mod/calculated_metrics/spec/set/metric_type/score_spec.rb
```diff
@@ -229,18 +229,35 @@ def answer metric
...
+    @metric = create_metric(name: "#{@metric_name}+Big Brother", type: :score,
+                            formula: '{"2":4, "3":6, "else": 5}')
+    expect(score_value("Sony Corporation", "2010")).to eq "5.0"
+  end
+
+  example "score unknown value", as_bot: true do
+    @metric_title = "RM"
+    @metric_name = "Joe User+RM"
+    @metric = create_metric(name: "#{@metric_name}+Big Brother", type: :score,
+                            formula: '{"Unknown":0, "else": 10}')
```
xithan, line 177: This test calculates the score formula that Laureen (?) asked for.
There is no way yet to change the interface to the translation editor but you can enter that formula above into the standard text formula editor to get that result. "else" is only the backend syntax. We can still call it "none of the above" in the interface.

mod/isic/lib/formula/isic.rb
```diff
@@ -0,0 +1,13 @@
+module Formula
+  # Calculator class for ISIC division and group codes
+  class Isic < HardCoded
```
xithan, line 3: Calculator no longer receives a card object. I created a new calculator classes for your hard coded cases to make them work again.
(and also fixed the bug that validation wasn't called)

mod/calculated_metrics/set/type_plus_right/metric/formula/descendant.rb
```diff
@@ -1,7 +1,7 @@
 
 def inheritance_formula
   item_names.map do |item|
-    "{{#{item}}}"
+    "{{#{item} | not_researched: false}}"
```
xithan, line 4: I decided to go for "false". Maybe more intuitive in a boolean expression. But nil should work, too.

mod/calculated_metrics/lib/formula/wolfram/unknowns.rb
```diff
@@ -1,36 +1,24 @@
...
-
-      def handle_unknowns company, year
-        @unknown_input ||= Hash.new_nested Array
-        catch(:unknown) do
+      # Deals with the "unknown: result_unknown" option for Wolfram formulas.
+      # If the formula is supposed to return "Unknown" because of an
+      # "Unknown" input value, then we have skip the calculation for that answer and
+      # add the "Unknown" after we got the result from the Wolfram server
+      def handle_unknowns input_values, company, year
+        @unknown_result ||= Hash.new_nested Array
```
xithan, line 10: You removed all this in your branch. I restored it. I didn't see a solution in your code for the problem describe in the comment above.

mod/calculated_metrics/lib/formula/parser.rb
```diff
@@ -0,0 +1,77 @@
+module Formula
+  class Parser
+    OPTIONS = %i[year company unknown not_researched].freeze
+    COUNT_RELATED_FUNC =
+      { "CountRelated" => "Total[ {{always one|company: Related" }.freeze
```
xithan, line 5: here is the CountRelated stuff

mod/calculated_metrics/lib/formula/calculator/input_item/options/unknown_option.rb
```diff
@@ -0,0 +1,48 @@
...
+          end
+
+          private
+
+          def interpret_unknown_option
+            case unknown_option
+            when "no_result"      then extend UnknownNoResult
+            when "result_unknown" then extend UnknownResultUnknown
+            else
+              extend UnknownPassValue
```
xithan, line 41: I haven't thought that through if this is dangerous. Maybe we should be more restrictive what we allow as custom values for "Unknown".

PR #934, opened 2019-01-04
lists in formula

xithan:
support wolfram list syntax in ruby formulas


mod/calculated_metrics/lib/formula/ruby.rb
```diff
@@ -8,7 +8,7 @@ module Formula
   # The formula may only consist of numbers and the symbols and functions
   # listed in SYMBOLS and FUNCTIONS
   class Ruby < NestFormula
-    SYMBOLS = %w{+ - ( ) [ ] . * , / || &&}.freeze
+    SYMBOLS = %w[+ - ( ) \[ \] . * , / || && { } ].freeze
```
ethn, line 11: Do `{` and `}` have the same meaning in Wolfram?

ethn (commented):
I want to be sure we're not reintroducing things that work in ruby but not Wolfram; that causes problems when people start with a working ruby formula and then add a wolfram function to it.

mod/calculated_metrics/lib/formula/ruby.rb
xithan: No, but now they have in Ruby the same meaning they have in Wolfram.

xithan:
We are introducing things that work in Wolfram but not in Ruby (see the specs  https://github.com/wikirate/wikirate/pull/934/files#diff-60fc292aa68bbfc8b8a1053254344584R73)

ethn (approved):

PR #935, opened 2019-01-09
fixed sql for company options

xithan:
To make the formula parsing and sql building less complicated I allow for company conditions either &&s or ||s but not both. I don't think that's a restriction anybody will notice in the near future or maybe ever and it saves me a lot of work.


mod/calculated_metrics/lib/formula/calculator/input_item/options/company_option/company_search/abstract_condition.rb
```diff
@@ -14,62 +14,75 @@ def initialize string, id
...
+                @object_sql ||= tablefy "object_company_id"
               end
 
               def metric_sql
                 tablefy "metric_id = #{@metric_id}"
               end
 
               def inverse?
-                @inverse
+                false
```
ethn, line 38: why?

ethn: oh, I think it's because this lookup always happens on the direct.  maybe a short comment?

mod/calculated_metrics/lib/formula/calculator/input_item/options/company_option/company_search/related_condition.rb
```diff
@@ -0,0 +1,102 @@
...
+              end
+
+              def inverse_metric_ids
+                @conditions.select(&:inverse?).map(&:metric_id)
+              end
+
+              def validate_logic
+                return unless @str.include?("&&") && @str.include?("||")
+
+                raise Condition::Error, "mix of '&&' and '||' is not supported, yet"
```
ethn, line 94: this seems like a reasonable policy for now.

mod/calculated_metrics/spec/lib/formula/calculator/input_item/company_option/company_query_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+                     UNION
+                     SELECT object_company_id as subject_company_id,
+                            subject_company_id as object_company_id,
+                            metric_id, inverse_metric_id, year, value
+                     FROM relationships
+                     WHERE inverse_metric_id = #{less_evil_id}
+                   ) AS r0
+                   WHERE ((r0.inverse_metric_id = #{less_evil_id})
+                   OR (r0.metric_id = #{supplied_by_id} && r0.value = "Tier 1"))
+      SQL
```
ethn, line 98: I'm not usually in favor of long exact string matching in these, but I agree that just testing results doesn't capture enough here.

mod/calculated_metrics/spec/lib/formula/calculator/input_item/company_option/company_query_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+  let(:supplied_by_id) { Card.fetch_id "Commons+Supplied by" }
+  let(:deadliness_id) { Card.fetch_id "Jedi+deadliness" }
+
+  describe "#sql", output_length: 5000 do
+    def sql str
+      described_class.new(str, nil).sql.tr("\n", " ").squeeze(" ").strip
+    end
+
+    def eq_sql sql
+      eq sql.strip_heredoc.tr("\n", " ").squeeze(" ").strip
```
ethn, line 15: wondering if we should `gsub(/\s+/, " ")` or something (on both sides) so changes in white space don't break tests.

ethn (approved):
cool stuff.

mod/calculated_metrics/spec/lib/formula/calculator/input_item/company_option/company_query_spec.rb
xithan: that's what `squeeze` does (and it's also called on the other side)

mod/calculated_metrics/lib/formula/calculator/input_item/options/company_option/company_search/abstract_condition.rb
xithan: there is an extra module for inverse relationships that overrides that method

PR #936, opened 2019-01-11
fix cache bug

xithan:
includes delayed job cache fix

### DIFF
--- a/mod/calculated_metrics/lib/formula/calculator.rb
+++ b/mod/calculated_metrics/lib/formula/calculator.rb
@@ -123,7 +123,7 @@ module Formula
     def replace_nests content=nil
       content ||= formula
       index = -1
-      content.gsub(/{{[^}]*}}/) do |_match|
+      content.gsub(/{{[^{}]*}}/) do |_match|
         index += 1
         yield(index)
       end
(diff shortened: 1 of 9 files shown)


mod/calculated_metrics/set/type/metric/calculations.rb
```diff
@@ -10,7 +10,9 @@
 end
 
 def calculation_in_progress!
-  Answer.where(id: all_dependent_answer_ids).update_all(calculating: true)
+  ids = all_dependent_answer_ids
+  Answer.where(id: ids).update_all(calculating: true)
+  Answer.where(id: ids).each(&:expire)
```
ethn: Aren't these answer objects and not cards?

ethn: Weird, Answer objects have an `expire` method, but I can't figure out where it's defined.

```
[5] pry(main)> Answer.first.method(:expire).source_location
=> nil
```

ethn (commented):
Can you explain what `expire` is getting called?  It looks like both times it's called the expiration is on answer objects.

mod/calculated_metrics/set/type/metric/calculations.rb
xithan: Answer objects delegate missing method calls to the answer card object.

ethn: Ahhhh.  thank you.

ethn (approved):

PR #938, opened 2019-01-15
bug fix soup

ethn:
(changes will be easier to see once we merge in the `update!` fixes.)


mod/calculated_metrics/set/type/metric/dependencies.rb
```diff
@@ -10,18 +10,22 @@ def each_dependent_metric
...
 end
 
 def dependency_tree
   DependencyTree.new directly_dependent_metrics
 end
 
 def score_metrics
-  Card.search type_id: MetricID, left_id: id
+  @score_metrics ||=
+    Card.search type_id: MetricID, left_id: id
```
xithan, line 25: Are you sure that this does no harm?  I've seen too often how this kind of caching blew up something

ethn: `Sure` would be an overstatement, but the only place this is called is in `directly_dependent_metrics`, which also calls `formula_metrics`, which already uses this kind of caching.  It seemed odd to me to use it in the one but not the other. I've also tried (and failed) to concoct a scenario when it would be a problem. But if those don't seem like reason enough, I can undo it.

ethn: verdict?

xithan: acceptable

xithan (approved):

PR #944, opened 2019-01-29
relationships can be added and edited

xithan:


mod/research/set/self/research_page.rb
```diff
@@ -15,6 +15,14 @@
     _render_core
   end
 
+  view :add_relation, cache: :never do
+    @answer_view = :research_form
+    @answer_card = Card.new name: [metric, company, year.to_s],
+                            type_id: RelationshipAnswerID
+    @answer_card.define_singleton_method(:unknown?) { true }
```
ethn, line 22: I gather `new?` won't work?

ethn: (maybe worth a comment?)

ethn (approved):
If this is really new functionality, then you pulled it off with impressively little new code.

mod/research/set/self/research_page.rb
xithan: That's one of the many routes I tried. Could be that it's no longer necessary. I have to check.

xithan:
The form and some coffeescript magic was already there for creating relationships. The problem was to make that form work for the case that a year has already relationships.

mod/metrics/set/abstract/value/edit.rb
```diff
@@ -28,7 +28,17 @@ def edit_fields?
...
 
   def value_field_card_and_options
-    [card, { title: "Answer", editor: :standard, show: :help }]
+    val_card =
+      if card.relationship?
+        # For adding new relationship answers.
+        # The card is the (existing) metric answer card M+C+Y+values with
+        # the company count.
+        # With this change we get the right editor for creating a new relationship value.
+        Card.new name: card.name, type_id: card.metric_card.value_cardtype_id
```
xithan, line 37: This did the trick: Replace the card object with a new card with the same name. But it needs a decko change to make the form prefixes work.

ethn: says 330 files changed in decko.  I'll wait for the decko PR :)

xithan: 330? I changed one! Did I accidentally merged the bridge stuff?

ethn: I assume so :).  You can back out or we can push ahead with the bridge, either way.

PR #945, opened 2019-02-02
project metrics and companies paging

ethn:


mod/projects/set/type_plus_right/project/metric.rb
```diff
@@ -42,17 +42,15 @@ def filter_card
   end
 
   view :core do
-    wrap_with :div, class: "progress-bar-table" do
-      metric_progress_table
-    end
+    nest Card.fetch(card.name, :project), view: :content, items: { view: :bar }
```
ethn, line 45: The point of all these bar-related changes is that by moving these into a separate card, we can automatically get paging, which is badly needed on these project pages.

mod/researched_metrics/set/type/relationship_answer.rb
```diff
@@ -131,8 +131,8 @@ def legend
                   value: card.value,
                   import: card.imported?,
                   comments: field_nest(:discussion, view: :core),
-                  subject_company: card.company,
-                  object_company: card.related_company
+                  subject_company: Card.fetch_name(card.company),
+                  object_company: Card.fetch_name(card.related_company)
```
ethn, line 135: hack to fix non-standard names in JSON.  Would prefer to pull these from relationships table.

mod/widgets/lib/open_corporates/api.rb
```diff
@@ -107,6 +107,6 @@ def api_key
     end
   end
 
-  class APIError < StandardError
+  class APIError < Card::Error::UserError
```
ethn, line 110: When these things are UserError's, NewRelic (et al) won't get notified when a user makes an API mistake.

mod/caching/set/abstract/answer_table_cached_count.rb
```diff
@@ -70,3 +70,8 @@ def wql_from_content
 def cache_query?
   false
 end
+
+# TODO: paging.  for now returns top 50.
+def item_cards _args={}
+  search.sort_by(&:name)[0, 50]
+end
```
ethn, line 77: temporary solution for very long page loads.

PR #950, opened 2019-02-27
starting in on context work with +answer harmonization.

ethn:


mod/caching/set/type_plus_right/metric/metric_answer.rb
```diff
@@ -1,12 +1,10 @@
 # cache # of answers for metric
-include_set Abstract::AnswerTableCachedCount, target_type: :answer
-
-def search_anchor
-  { metric_id: left.id }
-end
+include_set Abstract::CachedCount
```
ethn, line 2: This set is now merged with what used to be +all metric values.  So it's much simpler.

mod/filter/set/abstract/right_filter_form.rb
```diff
@@ -0,0 +1,5 @@
+format :html do
+  view :core, cache: :never do
+    filter_fields slot_selector: ".RIGHT-answer.filter_result-view"
+  end
+end
```
ethn, line 5: this wikirate-specific code was in decko before.

mod/metrics/lib/card/answer_query.rb
```diff
@@ -146,7 +142,7 @@ def run_filter_query
     def prepare_filter_args filter
       @filter_args = filter.deep_symbolize_keys
       # TODO: remove following (and handle consequences in tests!)
-      @filter_args[:latest] = true unless filter[:year] || filter[:metric_value]
+      # @filter_args[:latest] = true unless filter[:year] || filter[:metric_value]
```
ethn, line 145: This is in some ways the biggest change.  Latest is still the default filter in the main html views, but not down deep in AnswerQuery.

mod/metrics/spec/lib/card/fixed_company_answer_query_spec.rb
```diff
@@ -0,0 +1,372 @@
+require "./test/seed"
+
+RSpec.describe Card::FixedCompanyAnswerQuery do
```
ethn, line 3: these tests were moved from what is now the +answer card specs.  Feels right; they're really testing answer query code. (Also the other tests were getting long and complicated).

mod/researched_metrics/set/type_plus_right/metric/metric_answer.rb
```diff
@@ -1,13 +1,14 @@
...
 
 def query_class
   FixedMetricAnswerQuery
 end
 
-def default_sort_option
-  :value
+def filter_card_fieldcode
+  :metric_company_filter
 end
```
ethn, line 12: already the default.

mod/wikirate/lib/javascript/script_wikirate_coffee.js.coffee
```diff
@@ -25,3 +25,16 @@ appendParentToAddItem = (parent) ->
     anchor.attr "href", new_href
 
 
+$(window).ready ->
+  $("body").on "click", "a.card-paging-link", ->
+    id = $(this).slot().attr("id")
+    #unless history.state?
+    #  history.replaceState(slot_id: id, "")
+    history.pushState(slot_id: id, url: this.href, "", location.href);
```
ethn: If we can quickly do this on every slot update, we probably should.

xithan (approved):

PR #952, opened 2019-03-14
editing with rich headers

ethn:


mod/metrics/set/type/metric/listing.rb
```diff
@@ -2,6 +2,7 @@
   view :bar_left do
     render :thumbnail_with_vote
   end
+  view :bar_expanded_left, :bar_left
```
ethn, line 5: not loving having a dozen of these...

xithan (approved):

PR #958, opened 2019-03-30
bar and project site work

ethn:


mod/wikirate_assets/set/all/wikirate_shared.rb
```diff
@@ -0,0 +1,48 @@
...
+              "Legal" => "https://project.wikirate.org/Legal" }.freeze
+
+SECONDARY_MENUS = { "Legal" => ["Legal", "Privacy Policy", "Licensing", "Disclaimers",
+                                "Terms of Use", "Notice"],
+                    "Social" => %w[Facebook Twitter Instagram Telegram] }.freeze
+
+format do
+  def nav_menus
+    NAV_MENUS
+  end
```
xithan, line 29: If your constants have to be overwritable you can use the setting api
```ruby
format do
  setting :nav_menus
  nav_menus NAV_MENUS
end
```

xithan (approved):

mod/wikirate_assets/set/all/wikirate_shared.rb
ethn: Cool.  In this case it wasn't for overwriting; I just wanted a convenient way to pull the hash into haml templates.  I refer to it in three places, so the full constant name was unwieldy.

PR #964, opened 2019-04-13
 adding subproject metrics/companies to parent

ethn:


mod/projects/set/abstract/project_scope.rb
```diff
@@ -0,0 +1,105 @@
+# module for project-scope-definining cards
+# (+company, +metric, and +year on projects)
```
ethn, line 2: I'm not sure this is the best naming, but the idea was that there was a lot of duplicate code in +metric, +company, and +year, and I wanted to consolidate rather than add more.

mod/metrics/set/abstract/metric_answer/research.rb
```diff
@@ -82,18 +81,23 @@ def answer_delete_button
...
+                  class: "btn btn-sm btn-outline-danger pull-right",
                   'data-confirm': confirm,
                   "data-disable-with": "Deleting"
   end
 
   view :edit_buttons do
     class_up "form-group", "w-100 m-3"
     button_formgroup do
-      [standard_submit_button, standard_cancel_button, answer_delete_button]
+      [standard_save_button, cancel_research_button, answer_delete_button]
```
ethn, line 92: cancel needed to be changed to reload page; otherwise the right side did not correct itself.

xithan (approved):

PR #967, opened 2019-05-01
research page optimizations

ethn:


ethn:
I'm not sure it's actually all that much faster yet :(.  But I think the changes are generally positive.

mod/wikirate_csv_import/lib/csv_row/structure/answer_csv.rb
```diff
@@ -66,7 +66,7 @@ def check_existence_and_type name, type_id, type_name=nil
         end
       end
 
-      def resolve_source_duplicates existing_source_card
+      def resolve_source_duplication existing_source_card
```
ethn, line 69: please confirm that this change is correct.  `resolve_source_duplicates` is never called. Looked to me like this was intended to be an override.

mod/metrics/set/abstract/metric_answer/flags.rb
```diff
@@ -2,61 +2,54 @@ def imported?
...
-def commented?
-  disc = fetch trait: :discussion
-  disc&.content.present?
-end
-
 format :html do
-  view :flags, cache: :never do
-    output [checked_value_flag, comment_flag, imported_flag, calculated_flag]
+  view :flags do
+    output flags
```
xithan: the `output` is not necessary for arrays in html views

mod/metrics/set/abstract/metric_answer/flags.rb
```diff
@@ -2,61 +2,54 @@ def imported?
...
   end
 
-  view :small_flags, cache: :never do
-    output do
-      [:checked_value, :comment, :imported].map do |flag_name|
-        flag = send "#{flag_name}_flag"
-        "<small>#{flag}</small>"
-      end
+  view :small_flags do
+    output(flags.map { |flag| "<small>#{flag}</small>" })
```
xithan: the same here

mod/wikirate_csv_import/lib/csv_row/structure/answer_csv.rb
xithan: to me too

xithan (approved):

PR #968, opened 2019-05-08
merging fashion companies (+ decko fixes)

ethn:


mod/metrics/set/abstract/answer_search.rb
```diff
@@ -44,7 +44,7 @@ def item_view_from_query
     merge_filter_defaults
     wrap_with :div, class: "filter-form-and-result nodblclick" do
       class_up "card-slot", "_filter-result-slot"
-      output [_render_filter_form, _render_filter_result]
+      output [_render_filter_form, _render_filter_result(home_view: :filter_result)]
```
ethn, line 47: This is a hack to get around what I see as a decko bug in the handling of home views.  The filter result slot doesn't get :filter_result as its home view by default.

VERSION
```diff
@@ -1 +1 @@
-0.18.5
```
ethn, line 1: I tagged and released 0.18.6 between these two...

ethn:
The context name problems I had introduced were pretty ugly, and I needed to run the fashion merge script, so these got deployed to production quickly.  (the other two current PRs are not deployed yet)

xithan (approved):

PR #969, opened 2019-05-09
view permission tweaks

ethn:


mod/calculated_metrics/set/ltype_rtype/record/year.rb
```diff
@@ -2,10 +2,6 @@
 
 include_set Type::MetricAnswer
 
-def unknown?
-  answer.blank?
-end
-
```
ethn, line 8: not necessary, since `virtual?` is overridden.

I note this in some coming comments, but imo `virtual?` is the only state method that we should ever overwrite.

mod/metrics/set/abstract/value/events.rb
```diff
@@ -23,10 +23,11 @@
...
 end
 
-event :update_double_check_flag, :validate, on: [:update, :delete], changed: :content do
+event :reset_double_check_flag, :validate, on: [:update, :delete], changed: :content do
   [:checked_by, :check_requested_by].each do |trait|
-    next unless left.fetch trait: trait
-    attach_subcard name.left_name.field_name(trait), content: ""
+    full_trait_name = name.left_name.field_name trait
+    next unless Card.real?(full_trait_name) && !subcard(full_trait_name)
+    attach_subcard full_trait_name, content: ""
```
ethn, line 30: I'm not sure what change made this break (probably something to do with unknown?), but this was causing problems when the answer value and the double check field were updated in the same request.

mod/researched_metrics/set/right/checked_by.rb
```diff
@@ -106,21 +106,10 @@ def option_label_text _option_name
   end
 
   view :core, template: :haml
-
-  # FIXME: this view is wrongly cached if it's moved to a haml template
-  #   To see how it fails add `template: :haml` and run the double_check.feature
```
ethn, line 111: one FIXME for me ;)

I think my view handling updates cleaned this up. Seems to work fine now.

mod/widgets/set/right/wikipedia.rb
```diff
@@ -51,13 +56,10 @@ def jurisdiction_name oc_code
...
-  view :edit do
-    Card.exists?(card.name) ? super() : _render_new
-  end
 
-  def unknown_disqualifies_view? _view
-    false
-  end
+  # view :edit do
+  #   Card.exists?(card.name) ? super() : _render_new
+  # end
```
ethn, line 62: not confident here, but things seem to work better without this.

mod/wikirate_assets/spec/set/type/image_spec.rb
```diff
@@ -1,27 +1,19 @@
...
-                                image: @file
-      missing_image = image_card.format.subformat(@missing_card)._render_core
-
-      image_card_format = image_card.format
-      image_card_format.instance_variable_set "@denied_view", :core
+    it "is NOT wrapped when rendered from core" do
+      expect(new_image_view(:core)).not_to match(/missing-view/)
+    end
 
-      expect(image_card_format.render_missing).to eq(missing_image)
```
ethn, line 16: This really seemed like a lot of hullaballoo for testing very little.  Am I missing something?

mod/calculated_metrics/set/ltype_rtype/record/year.rb
ethn: I guess it's possible that there could be a case where there is a real card but the lookup answer is blank.  But that would be a data problem.

xithan: It was already common practice to override `virtual?` when I did this. I don't remember the details but I think there was a reason that I did it this way. 
Could be to keep the edit menu because all these cards have editable content. Maybe that is no longer necessary because the old menu logic is gone.

mod/wikirate_assets/spec/set/type/image_spec.rb
xithan: probably very old

xithan (approved):

PR #979, opened 2019-06-03
thumbnail upgrade / simplification

ethn:


mod/layout/set/abstract/thumbnail.rb
```diff
@@ -1,6 +1,7 @@
 format :html do
-  view :thumbnail_plain do
-    wrap_with :div, thumbnail_content, class: flex_css
```
ethn, line 3: I couldn't figure out enough of a difference between `thumbnail` and `thumbnail plain` to document it, so I merged them.  I did see that thumbnail didn't have all the "data" in the div, but that hardly seemed worth a separate view.

mod/caching/set/abstract/tagged_by_cached_count.rb
```diff
@@ -22,8 +24,18 @@ def self.included host_class
...
+    end
+
+    define_method :right_plus_val do
+      [Card::Codename.id(host_class.tag_pointer), { refer_to: left.id }]
     end
   end
 end
+
+# FIXME: hack. otherwise filter wql can overwrite right_plus
+def filter_wql
```
ethn, line 37: I don't love this.  Would love to discuss longer term plans for wql/cql that include solutions for the unique key problem.

mod/projects/set/type_plus_right/pointer/project.rb
```diff
@@ -4,7 +4,7 @@
 include_set Type::SearchType
 
 def virtual?
-  true
+  new?
```
ethn, line 7: I think this is a better practice than `true`. It avoids ever returning true when there is a real card.

mod/research/set/self/research_page/metric_slot.haml
```diff
@@ -9,8 +10,7 @@
     %div
       .metric-color.d-flex.flex-nowrap.align-items-center
         .image-box.icon
-          %span.img-helper
```
ethn, line 12: I couldn't figure out what all these image helpers were supposed to do.

PR #983, opened 2019-06-08
new details handling

ethn:


mod/caching/set/type_plus_right/metric/source.rb
```diff
@@ -15,7 +15,7 @@ def cache_query?
 end
 
 def answer_ids
-  Answer.where(metric_id: left.id).where.not(answer_id: :nil).pluck :answer_id
+  ::Answer.where(metric_id: left.id).where.not(answer_id: :nil).pluck :answer_id
```
xithan: Is colon nil really the right syntax?

mod/layout/lib/javascript/script_table.js.coffee
```diff
@@ -1,96 +1,77 @@
...
-  no_toggle = null
-  $('body').on 'click', ".details-toggle  .details", ->
-    no_toggle = true
+  $('body').on 'click', '.details ._update-details', (e) ->
+    (new decko.details).add $(this)
+    e.preventDefault()
 
-  $('body').on 'click', ".details-close-icon", ->
-    trToggleDetails this
+decko.details = (dSlot) ->
```
xithan, line 14: a new practice to organize coffeescript?

xithan (approved):

mod/caching/set/type_plus_right/metric/source.rb
ethn: ooh, I don't think so.  I suspect this is a(n old) bug.

mod/layout/lib/javascript/script_table.js.coffee
ethn: Taking an object oriented approach to decko.filter made things _much_ cleaner.  I went with the same approach this time, and I couldn't swear the rewards were as high, but I do find it easier to read.

The other nice thing is that when you do this (and put very little logic in the event handling), it's easier to test in the browser, because `decko` is global.

PR #988, opened 2019-07-02
range chart fixes

ethn:


mod/metrics/lib/card/vega_chart/range_chart.rb
```diff
@@ -41,29 +40,16 @@ def data_item_hash filter, _count
...
-      #
-      # def x_labels
-      #   precision = 3
-      #   while precision < 10
-      #     human = @labels.map do |num|
-      #       @format.number_to_human num.to_f, precision: precision
-      #     end
-      #
-      #     return human if human.uniq.size == human.size
-      #     precision += 1
```
xithan, line 64: What's the problem with this?

ethn: The old issue was that when used with logarithmic scales, it sometimes produced non-unique labels, and for whatever reason, vega reduced it to the uniques.

The above fixed it, but `format: "~s"` fixed it much more easily.

PR #993, opened 2019-07-12
option heritage fix

xithan:


mod/metrics/set/abstract/metric_thumbnail.rb
```diff
@@ -30,7 +30,7 @@ def fixed_thumbnail_subtitle
 
   def formula_options
     %i[year company unknown not_researched].map do |key|
-      next unless (value = voo.heritage_options(key))
+      next unless (value = voo.live_options[key])
```
ethn, line 33: or `voo.send(key)`.  I suppose this is better.

ethn (approved):

PR #1039, opened 2020-02-21
filtering on relationship answers

ethn:


mod/calculated_metrics/spec/lib/formula/wolfram_spec.rb
```diff
@@ -5,7 +5,7 @@
   include_context "with calculator stub"
   include_context "company ids"
 
-  it "simple formula" do
+  xit "simple formula" do
```
ethn, line 8: wolfram maxed out for month.  commenting for now...

PR #1087, opened 2020-09-03
Move layout to card-mods

GerryG:
Although I did not remove it from the mod directory, the dependencies in Gemfile and Modfile are now the card-mod-layout gem.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,6 +9,13 @@ PATH
   specs:
     card-mod-csv_import (0.8)
       card
+      card-mod-layout
+
+PATH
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -9,6 +9,13 @@ PATH
   specs:
     card-mod-csv_import (0.8)
       card
+      card-mod-layout
```
ethn, line 12: Semaphore is saying you need to update Gemfile.lock. Looks like you already did, but maybe try again?

ethn:
I have to do a little more rearranging before we do this.

PR #1174, opened 2021-02-25
year option "add" and "latest"

xithan:


mod/calculated_metrics/lib/formula/calculator/input_item/options/year_option.rb
```diff
@@ -55,6 +57,10 @@ def apply_year_option value_data, year
               ip.map { |y| value_data[y] }
             when Proc
               ip.call(year).map { |y| value_data[y] }
+            when :all
+              value_data.values
+            when :latest
+              value_data.values.last
```
ethn, line 63: nice! I was hoping it would be this straightforward.

This method is making code climate unhappy; I usually want to fix the code complexity ones. (less concerned about 251 lines of code in formula_spec.rb)  We could do something like:

```
  send "apply_#{ip.class}_year_option", year, ip
```

and then define, eg, `apply_integer_year_option`...

ethn (commented):
I think we should fix the apply_year_option complexity issues in code climate. but otherwise this looks great.

mod/calculated_metrics/lib/formula/calculator/input_item/options/year_option.rb
ethn: I'm working on this now.

PR #1175, opened 2021-02-26
gem updates and year options

ethn:


mod/calculated_metrics/lib/formula/calculator/input_item/options/year_option.rb
```diff
@@ -45,18 +47,37 @@ def process_year_option
           #   for every year
           # @param the year we want the input data for
           def apply_year_option value_data, year
-            year = year.to_i
-
             ip = processed_year_option
+            method = "apply_#{ip.class.to_s.downcase}_year_option"
+            illegal_input_processor! ip unless respond_to? method
+
+            send method, value_data, ip, year.to_i
```
xithan, line 54: Not worth to fight about this, but I don't believe Codeclimate, that this is easier to read than the case-statement.
But I do think that something here is conceptually wrong. There shouldn't be a case statement necessary at all. It should be handled in the specialised modules year_all/year_latest/year_list/...
Seems wrong that all what year_all does, is return :all and then we have to detect that here and do the real work. But I don't see the solution.

xithan (approved):

mod/calculated_metrics/lib/formula/calculator/input_item/options/year_option.rb
xithan: Here is an argument: Your code is technically still a case-statement but it's obfuscated. It's harder to see what the possible cases are and what the differences are.

ethn: Yeah, I think I've been sucked into codeclimate as a bit of a game. On the whole it's pretty clearly beneficial, but there are probably some cases where the gain is negligible or there may even be a minor cost.

Fwiw, more often than not, I find things easier to debug when they're broken down this way. You get more info in backtraces, idea steps, etc.  And the larger the developer community is, the more efficient it is to say "defer to codeclimate" rather than debating code aesthetics.  We're not large yet, obviously, but someday...

PR #1178, opened 2021-03-04
Introduce region cards

xithan:


mod/profiles/set/self/wikirate_company/factories.rb
```diff
@@ -59,11 +59,12 @@ def search_factory_ids
             return: :id }
     if country_code
       cql[:right_plus] = [{ codename: "headquarters" },
-                          { refer_to: { codename: country_code.to_s } }]
+                          { refer_to: { id: country_code.to_s } }]
```
ethn: If the `to_s` is actually doing something, we may need a comment, because it's hard to see why it's needed (or even why it was needed before).  Otherwise let's get rid of it.

mod/profiles/set/self/wikirate_company/factories.rb
```diff
@@ -59,11 +59,12 @@ def search_factory_ids
...
     end
     Card.search cql
   end
 
+  # card id of a region card (Used to be the OC jurisdiction code)
   def country_code
```
ethn: maybe we should change the method name to `region_id`? We don't really have to change the params; the comment will suffice.

db/migrate_cards/20210225204833_jurisdiction_to_region.rb
```diff
@@ -0,0 +1,35 @@
...
+    ensure_card name: ["Region", :type, :structure], type_id: NestListID,
+                content: "{{+Continent|labeled|name}}\n" \
+                         "{{+OpenCorporates Jurisdiction key|labeled|name}}\n" \
+                         "{{+ILO Region|labeled|name}}\n" \
+                         "{{+Country|labeled|name}}"
+    Card::Cache.reset_all
+
+    Card.search(type_id: Card::JurisdictionID) do |card|
+      ensure_card name: "#{card.name}+OpenCorporates Jurisdiction key", content: card.codename,
+                  type_id: Card::JurisdictionID
```
ethn: This wasn't in the spec, but it occurs to me that while we're doing this, we should go ahead and add a country code as an identifier for each region that is a country. In other words if the jurisdiction key is just two characters (not including the oc_ prefix) then we should add a `+Country Code` card with that value.

I'm pretty confident this will prove useful for data integration some day.

db/migrate_cards/20210225204833_jurisdiction_to_region.rb
```diff
@@ -0,0 +1,35 @@
...
+    ensure_card name: "Region", type_id: CardtypeID, codename: "region"
+    ensure_card name: ["Region", :type, :structure], type_id: NestListID,
+                content: "{{+Continent|labeled|name}}\n" \
+                         "{{+OpenCorporates Jurisdiction key|labeled|name}}\n" \
+                         "{{+ILO Region|labeled|name}}\n" \
+                         "{{+Country|labeled|name}}"
+    Card::Cache.reset_all
+
+    Card.search(type_id: Card::JurisdictionID) do |card|
+      ensure_card name: "#{card.name}+OpenCorporates Jurisdiction key", content: card.codename,
```
ethn: Is there any good reason to keep the `oc_` prefix in the card content?  Shouldn't we sub it out now so we don't have to do it later?

mod/widgets/set/right/headquarters.rb
```diff
@@ -31,13 +31,13 @@ def needs_oc_mapping?
 # TODO: reduce duplicated code
 def jurisdiction_name oc_code
   oc_code = "oc_#{oc_code}" unless oc_code.to_s.match?(/^oc_/)
-  return unless Card::Codename[oc_code.to_sym]
-  Card.fetch_name oc_code.to_sym
+  return unless (card = Card.search(type_id: Card::JurisdictionID, content: oc_code)&.first)
```
ethn: To optimize this, we could build and cache a lookup hash `{ code1 => regionid_1, ... })`. It would only take one query.

Do we still need a juridiction type? Our more standard pattern for this is `right/jurisdiction`, no?

ethn (changes requested):
I asked for a little extra (out of spec) functionality, but this is very close.

It would be great also to add a single-use script for importing the values Laureen sent

PR #1225, opened 2021-06-28
ruby 3 support

ethn:


ethn:
still can't run wikirate in ruby 3, thanks to PDFKIT

PR #1412, opened 2022-06-29
definition of suggest_contexts method

vasgat:
I have made some changes into the open_search.rb to boost all types except sources for the autocomplete suggestions

### DIFF
--- a/mod/deckorate_search/set/self/search/open_search.rb
+++ b/mod/deckorate_search/set/self/search/open_search.rb
@@ -83,7 +83,7 @@ format do
   def suggestion_query
     { autocomplete: { prefix: search_keyword,
                       completion: { field: "autocomplete_field",
-                                    contexts: { type_id: filter_type_ids } } } }
+                                    contexts: { type_id: suggest_contexts } } } }
   end
 
   # constructs the type filtering clause for the os_query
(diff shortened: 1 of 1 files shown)


mod/deckorate_search/set/self/search/open_search.rb
```diff
@@ -11,7 +11,7 @@ def open_search_client
 # @param query [Hash] query hash
 # @result [Hash] ruby translation of JSON results
 # note: options configured in config/application.rb
-def search parameters={}
+def search parameters = {}
```
ethn, line 14: actually in method parameters we use the no-space convention.

mod/deckorate_search/set/self/search/open_search.rb
```diff
@@ -92,4 +92,14 @@ def os_type_filter
 
     yield[:filter] = { term: { type_id: type_param.card_id } }
   end
+
+  # constructs the context filtering clause for the suggest_query
+  # in case of multiple contexts we are favoring other contexts than source
+  def suggest_contexts
+    return filter_type_ids unless filter_type_ids.length() > 1
```
ethn, line 99: can you add the code suggested in slack and then make a new PR to main?

(I want to see if that triggers the code climate analysis.  I don't like merging in code that hasn't been through that).

ethn (changes requested):

PR #1413, opened 2022-06-29
definition of suggest_contexts method

vasgat:
I have made some changes into the open_search.rb to boost all types except sources for the autocomplete suggestions


mod/deckorate_search/set/self/search/open_search.rb
```diff
@@ -92,4 +92,12 @@ def os_type_filter
 
     yield[:filter] = { term: { type_id: type_param.card_id } }
   end
+
+  def suggest_contexts
+    return filter_type_ids unless filter_type_ids.length() > 1
+    contexts = []
+    filter_type_ids.each { |card_id| contexts.append({ "context": card_id, "boost": 2 }) unless card_id == 629 }
+    contexts.append({ "context": 629 })
+    return contexts
```
ethn: 1. we never use card ids in code.  instead use :mycodename.card_id

ethn: (in my local copy, the same card will have a different id but the same codename)

ethn: the PR automatically gets run through code climate, which helps enforce our style rules. For example, it will tell you: - line 99 is too long
- you don't need the "return" on line 101.
- you don't need the curlies on line 100.

etc.

ethn: I can't really read this code and understand what it's doing.  Why would you add all of them but 629 and then add 629?  Is it an order thing?

Refactoring might fix it, but otherwise we'd probably need a comment.

ethn (changes requested):

ethn:
let's just leave the one PR open.  closing duplicate.

PR #1517, opened 2023-07-06
changes on deckorate_graphql module

vasgat:
- add filtering on different wikirate entities


mod/deckorate_graphql/lib/graph_q_l/types/query.rb
```diff
@@ -3,16 +3,27 @@ module Types
...
 
           define_method(fieldname) { |**mark| ok_card codename, **mark }
-          define_method plural_fieldname do |name: nil, limit: 10, offset: 0|
-            card_search name, codename, limit, offset
+          define_method plural_fieldname do |limit: 10, offset: 0, **filter|
+            card_search codename, limit, offset, filter
+          end
+        end
+
+        def subcardtype_field fieldname, type, codename = nil
```
ethn: this seems like it's identical to the cardtype field, no?  The only difference is that it won't handle the singular variant.

My guess is that there's not really a need for this method. But if there is, it should be articulated in a comment, eg above the method.

mod/deckorate_graphql/lib/graph_q_l/types/answer.rb
```diff
@@ -5,10 +5,9 @@ class Answer < Card
       field :year, Integer, null: false
       field :company, Company, null: false
       field :metric, Metric, null: false
-
-      field :value, String, null: false
+      field :value, AnswerValue, null: false
       field :comments, String, null: true
-      field :sources, [Source], null: false
+      subcardtype_field :source, Company, :source
```
ethn: Why is the type company here?

mod/deckorate_graphql/lib/graph_q_l/types/answer_value.rb
```diff
@@ -0,0 +1,37 @@
...
+          input_value
+        elsif input_value.to_s.include? ","
+          input_value.split(",")
+        else
+          input_value
+        end
+      end
+
+      def self.coerce_result(ruby_value, context)
+        begin
```
ethn, line 52: This looks very odd. Can you explain what's going on here in a code comment?

ethn: I would recommend writing tests for this method to make sure it's behaving the way you want it to.

mod/deckorate_graphql/lib/graph_q_l/types/answer_value.rb
```diff
@@ -0,0 +1,37 @@
+module GraphQL
+  module Types
+    class AnswerValue < BaseScalar
+      description "An answer value that can be either a string, an integer, a float or an array"
+
+      def self.coerce_input(input_value, context)
+        if input_value.is_a?(Float) or input_value.is_a?(Integer) or input_value.is_a?(BigDecimal) or input_value.is_a?(BigInt)
```
ethn: another way to write this is:

```
case input_value
when Float, Integer, BigDeckomal, BigInt
  input_value
else
  ...
```

mod/deckorate_graphql/lib/graph_q_l/types/answer_value.rb
```diff
@@ -0,0 +1,37 @@
...
+          input_value
+        elsif input_value.to_s.include? ","
+          input_value.split(",")
```
ethn: I don't know if it's ever possible that input_value is not a string (and does not respond to split) but `input_value.to_s` includes ", ". If so this would fail

mod/deckorate_graphql/lib/graph_q_l/types/company.rb
```diff
@@ -2,18 +2,16 @@ module GraphQL
   module Types
     # Company type for GraphQL
     class Company < Card
-      field :answers, [Answer], "answers to questions about company", null: false do
-        argument :metric, String, required: false
-      end
+      subcardtype_field :answer, Answer, :metric_answer
```
ethn: This seems a little funny. Normally `subcardtype_field` does a Card search, and card searches on answers are going to be very slow. Here we want to use a lookup search.

But the method "answers" below actually overrides the `answers` method defined by `subcardtype_field` that calls `card_search`, right? so what does it do?

mod/deckorate_graphql/lib/graph_q_l/types/query.rb
ethn: After studying this further, here's what I would say:
1. I don't think defining the singular and plural variants in one method makes sense. If we need them both, we can specify both.
2. I'm not sure where the singular variant is used, and as written it doesn't look like it would ever do anything useful
3. So I don't think `subcardtype_field` is useful, and if it has a use, the name doesn't help me understand it.
4. I don't think there's anything deckorate-specific here; this can all be in card-mod

mod/deckorate_graphql/lib/graph_q_l/types/query.rb
```diff
@@ -23,79 +34,78 @@ def singular_field fieldname, type
...
+        end
+
       end
 
       cardtype_field :research_group, ResearchGroup
       cardtype_field :company_group, CompanyGroup
       cardtype_field :company, Company, :wikirate_company
+      cardtype_field :metric, Metric, :metric
+      cardtype_field :answer, Answer, :metric_answer
+      cardtype_field :relationship, Relationship, :relationship_answer
```
ethn: We don't want these to be cardtype_fields.  But we could add something called lookup fields.  Arguably the code for that should live in a module in the lookup mod.

mod/deckorate_graphql/lib/graph_q_l/types/metric_category_filter_type.rb
```diff
@@ -0,0 +1,9 @@
+module GraphQL
+  module Types
+    class MetricCategoryFilterType < BaseEnum
+      ::Card.fetch(:metric_type_type).item_cards.each do |item|
+        value item.card.name.url_key, value: item.card.name
+      end
```
ethn: all these filter types are just using the same code over and over again. We could do this much more efficiently.

ethn: Also, there's code in the filters for each of these.

For example, this one is in `filter_metric_type_options` in 05_metric_search.rb. That's where these should be coming from.

mod/deckorate_graphql/lib/graph_q_l/types/company.rb
```diff
@@ -2,18 +2,16 @@ module GraphQL
...
       field :logo_url, String, "url for company logo image", null: true
 
-      def answers metric: nil
-        query = { company_id: object.id }
-        query[:metric_id] = metric.card_id if metric
-        ::Card::AnswerQuery.new(query, {}, limit: 10).run
+      def answers metric_id: nil, limit: Card.default_limit, offset: Card.default_offset, **filter
+        filter[:company_id] = object.card_id
+        filter[:metric_id] = metric_id if metric_id
+        ::Card::AnswerQuery.new(filter, {}, limit: limit, offset: offset).lookup_relation.all
```
ethn: I don't think this is what we want to do.  

The general pattern of cardtype fields is something like (instance)+(cardtype). So in this case, it might be something like `Apple Inc+answers`.  So we could generally get that card by doing something like `[object, :metric_answer].card`.  If we do it like that, then it automatically handles `filter[:company_id] = object.card_id`; that's already in the default query. If we do it this way, we have to redefine the `answers` method everywhere we have an answers field.

ethn: So, this is rough pseudocode, but it might look something like:
```
def lookup_search search_card, limit=10, offset=0, **filter
  query_hash = (filter || {}).merge search_card.query_hash
  search_card.query.new(query_hash, limit: limit, offset:offset).run
end
```

Then you can just call something like `lookup_search [object, :metric_answer].card`.

mod/deckorate_graphql/lib/graph_q_l/types/dataset.rb
```diff
@@ -2,26 +2,26 @@ module GraphQL
...
-      def companies
-        object.wikirate_company_card.item_cards
-      end
-
-      def metrics
-        object.metric_card.item_cards
-      end
+      subcardtype_field :company, Company, :wikirate_company
+      subcardtype_field :metric, Metric
+      subcardtype_field :answer, Answer, :metric_answer
```
ethn: Again, these have to be written differently. They should be lookup table searches, not card table searches.

That's the reason they were written so awkwardly before, but we should be able to generalize it well.

mod/deckorate_graphql/lib/graph_q_l/types/query.rb
```diff
@@ -23,79 +34,78 @@ def singular_field fieldname, type
...
+                argument filter, MetricCategoryFilterType, required: false
+              elsif filter == :company_category
+                argument filter, CompanyCategoryFilterType, required: false
+              elsif filter == :company_group
+                argument filter, CompanyGroupFilterType, required: false
+              elsif filter == :designer
+                argument filter, MetricDesignerFilterType, required: false
+              elsif filter == :wikirate_topic
+                argument filter, TopicFilterType, required: false
+              else
```
ethn: I think would generalize this by making sure all the filter types map directly to the filter keys and then doing something like

```
filter_type = const_get_if_defined const("#{filter.to_s.camelize}FilterType") || String
argument filter, filter_type, required: false
```

I think you could get rid of almost all of this that way.

ethn (changes requested):
I know there's a lot here. We can set up a time to pair if that's best.

mod/deckorate_graphql/lib/graph_q_l/types/company_category_filter_type.rb
```diff
@@ -0,0 +1,9 @@
+module GraphQL
+  module Types
+    class CompanyCategoryFilterType < FilterType
+      ::Card.fetch(:commons_company_category).value_options_card.item_names.each do |option|
+        value option.url_key, value: option
```
ethn, line 6: use `filter_company_category_options`

mod/deckorate_graphql/lib/graph_q_l/types/deckorate_fields.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+      end
+
+      def plural_field fieldname, codename, type
+        field fieldname, [type], null: false do
+          argument :limit, Integer, required: false
+          argument :offset, Integer, required: false
+
+          codename.card.format.filter_keys.each do |filter|
+            filter_type = ENUM_FILTERS.include?(filter) ? GraphQL::Types.const_get("#{filter.to_s.camelize}FilterType"): String
```
ethn: I think the ENUM_FILTERS are the only part of this that are wikirate-specific, right? And I'm not sure it's necessary. You could just test whether the constant exists, no? (with `const_defined`)

If we did that, and if `card_field`, `cardtype_field`, `singular_field`, and `plural_field` were in the graphql mod, then I think they could be used by non-deckorate projects. And if lookup_field were in mod lookups (perhaps in a module named something like `Card::LookupGraphqlField`) then that could be used in non-deckorate projects with lookup fields.

I think you were trying to explain to me that you thought there was a reason it shouldn't be generalized, though? Can you explain?

mod/deckorate_graphql/lib/graph_q_l/types/wikirate_card.rb
```diff
@@ -0,0 +1,10 @@
+module GraphQL
+  module Types
+    class WikirateCard < Card
+      include DeckorateSearch
+      class << self
+        include DeckorateFields
+      end
```
ethn: You can just use `extend DeckorateFields`, and then you don't need `class << self`

ethn (changes requested):
This is looking great. Please make sure it passes all the tests (including the code climate style tests) before merging.

mod/deckorate_graphql/lib/graph_q_l/types/company_category_filter_type.rb
vasgat: There is an issue with the company categories on the test data. I am not sure why but it seems the content of the categories are something like A+A and I get the following error, if I use the generalize method: `filter_option_values(:metric, "company_category")` we developed: `Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but 'A+A' does not`

ethn (changes requested):
Please replace this PR with a request to the main branch. That will trigger all the testing that we want. 

It looks to me like an rspec is still failing.

mod/deckorate_graphql/lib/graph_q_l/types/company_category_filter_type.rb
ethn: `:commons_company_category.card` is more concise and easier to read than `::Card.fetch(:commons_company_category)`

mod/deckorate_graphql/lib/graph_q_l/types/deckorate_fields.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      def cardtype_field fieldname, type, codename = nil, is_card = false
+        codename ||= fieldname
+        plural_fieldname = fieldname.to_s.to_name.vary(:plural).to_sym
+        plural_field plural_fieldname, codename, type
+
+        define_method plural_fieldname do |limit: 10, offset: 0, **filter|
+          card_search codename, limit, offset, is_card, filter
+        end
+      end
```
ethn, line 29: did we decide not to move card_field and cardtype_field to the more general graphql module?

ethn:
@vasgat, whenever you're finished with the comments here, feel free to close this PR

ethn:
closing this one to avoid duplicate test runs on semaphore. The comments are, of course, still available if useful even if the PR is closed.

PR #1521, opened 2023-07-21
Relationship search standardization

ethn:


mod/deckorate_core/set/abstract/relationship_search.rb
```diff
@@ -0,0 +1,62 @@
...
+    ::Relationship.csv_titles detailed?
+  end
+
+  view :core do
+    detailed = detailed?
+    query.lookup_relation.map { |row| row.csv_line detailed }
+  end
+
+  def default_limit
+    nil
```
vasgat, line 60: Why the default_limit here is nil? My understanding is that it overwrites the default_limit from the lookup_search where the default limit is: 
`Auth.signed_in? ? 5000 : 500`

vasgat (approved):
I think the changes make sense. I left only one comment but probably is not something you need to address but rather me to understand.

PR #1526, opened 2023-08-23
add styling to attribution alert

cschweipert:


mod/deckorate_attribution/assets/style/reference.scss
```diff
@@ -1,3 +1,24 @@
 .TYPE-reference .RIGHT-Xsubject.card-editor {
   display: none
-}
\ No newline at end of file
+}
+
+.modal-header {
+  display: none;
+}
```
ethn, line 7: Wikirate makes heavy uses of modals. This will make it so that the modal header doesn't appear anywhere on the site. In general, it's best not to add styles to a selector you didn't add yourself unless you have a very good sense of all the places the selector is used. (Although, of course, you can use that broader selector along side of your own selector).

In this case, for now you can just leave the header there, even if it's a little ugly and not what the mockup says. I will look into ways to address that later.

mod/deckorate_attribution/assets/style/reference.scss
```diff
@@ -1,3 +1,24 @@
...
+
+.link-to-guide {
+  font-size: small;
+  margin-left: auto!important;
+  text-decoration: underline;
+}
+
+.custom-text {
+  font-weight: bold;
+}
```
ethn, line 17: "custom text" is a little too generic.   But there is also a bootstrap class that can help here: `fw-bold` (see https://getbootstrap.com/docs/5.0/utilities/text/)

> There are only two hard things in Computer Science: cache invalidation and naming things.
> 
> -- Phil Karlton

mod/deckorate_attribution/assets/style/reference.scss
```diff
@@ -1,3 +1,24 @@
...
+
+.alert_message {
+  align-self: center;
+  flex-direction: column;
+  margin-right: auto;
+  padding-left: 0.5rem;
+}
```
ethn, line 24: also a little too generic. "attribution-alert-message" perhaps?  (hyphens are a bit more idiomatic in CSS and underscores are in ruby)

I would especially recommend using bootstrap for padding, because those classes are responsive. Eg `ps-2` will use a little less space in mobile.

mod/deckorate_attribution/assets/style/reference.scss
```diff
@@ -1,3 +1,24 @@
...
+
+.modal-header {
+  display: none;
+}
+
+.link-to-guide {
+  font-size: small;
+  margin-left: auto!important;
+  text-decoration: underline;
+}
```
ethn, line 13: I think maybe all of this can be done with bootstrap classes.  If not, can we come up with a narrower (attribution-specific) name?

ethn (changes requested):
Great start! A couple of changes requested inline.

mod/deckorate_attribution/assets/style/reference.scss
cschweipert: Thanks! I had tried font-weight-bold and that didn't work. fw-bold works!

mod/deckorate_attribution/assets/style/reference.scss
cschweipert: ml-auto doesn't seem work with inline bootstrap but I was able to get it to work without the !important flag.

mod/deckorate_attribution/assets/style/reference.scss
ethn: I think that changed from Bootstrap 4 to 5. There were actually _lots_ of changes. Every now and then I try something that isn't working and I realize I'm using Bootstrap 4 docs.

mod/deckorate_attribution/assets/style/reference.scss
ethn: This might also be bootstrap 4.  In bootstrap 5 they largely switched from L and R (left and right) to S and E (start and end)

ethn:
In general, let's not mark conversations as resolved until we have pushed the code that resolves them. That makes tracking solutions much easier for reviewers.

PR #1529, opened 2023-08-23
refactor attribute alert message styling

cschweipert:
It is a more elegant to include title and body in one div as suggested by Ethan.


ethn (approved):
Even better!

Thank you!!

PR #1534, opened 2023-08-30
Fix clipboard bug, add close alert banner func, style copy button

cschweipert:


ethn (approved):
This all looks great to me. Thanks, Christina!

PR #1538, opened 2023-09-22
Clipboard

cschweipert:
Preserve Hyperlinks in Rich Text and copy the right text format to clipboard.


mod/deckorate_attribution/assets/script/clipboard.js.coffee
```diff
@@ -1,9 +1,10 @@
...
-    navigator.clipboard.writeText(clipboardContent)
+    content = $(this).closest(".tab-pane.active").find("._clipboard").html()
+    blob = new Blob([content], { type: "text/html" });
+    richTextInput = new ClipboardItem({ "text/html": blob });
+    navigator.clipboard.write([richTextInput])
       .then ->
-        console.log "Text copied to clipboard: #{clipboardContent}"
+        console.log "Text copied to clipboard: #{content}"
       .catch (error) ->
         console.error "Copy to clipboard failed:", error
```
ethn, line 10: `console.log` is for debugging and generally shouldn't be in production code. Perhaps we can comment this out?

mod/deckorate_attribution/assets/script/clipboard.js.coffee
```diff
@@ -1,9 +1,10 @@
 $ ->
   $("body").on "click", ".copy-button", ->
-    clipboardContent = $("#clipboard").text()
-    
-    navigator.clipboard.writeText(clipboardContent)
+    content = $(this).closest(".tab-pane.active").find("._clipboard").html()
```
ethn, line 3: Is this change from `text()` to `html()` what fixed the hyperlinks? Or was the Blob thing also necessary?

Do you get different behavior copying from the html and the rich text tabs? What's different?

ethn (changes requested):
Great work!  

Normally we'd want tests before approving new functionality, but since this is urgent and we're not really set up for javascript unit tests, let's go ahead and merge this in. 

I just have a couple of questions and a request to remove debugging (see below).

PR #1541, opened 2023-09-26
align close icon with hamburger

cschweipert:


ethn:
Can you pose the size question to melanie in PT? If you need a smaller font-size, you can override it on the `i` tag as standard css.

mod/deckorate_utility/set/all/wikirate/icons.rb
```diff
@@ -38,7 +38,8 @@
   board: :build,
   reference: :attribution,
   history: :history,
-  clipboard: :content_copy
+  clipboard: :content_copy,
+  close: :close
```
ethn, line 42: close is actually already defined in vendor/decko/mod/bootstrap/set/all/bootstrap/icon.rb

(it's used in decko, not just wikirate)

ethn (changes requested):
we're still waiting for the sizing fix, right?

ethn (approved):

PR #1542, opened 2023-09-26
Clipboard

cschweipert:
I kept the custom scss for the border radius to overwrite the bootstrap btn and btn-primary because the top and bottom right corners have no border radius.


mod/deckorate_attribution/set/type/reference/attribution_box.haml
```diff
@@ -1,10 +1,8 @@
 .d-flex.mb-3.align-items-stretch
   .p.clipboard._clipboard.p-2.rounded-start.flex-fill
     = content
-  .button.small.copy-button.d-flex.p-2.rounded-end
-    .d-flex.align-items-center.fw-bold
+  %button.copy-button.btn.btn-primary{ type: "button", "data-bs-toggle": "button" }
```
ethn, line 4: didn't "rounded-end" do the same thing the border-radius css does? The advantage is that if we change the default border-radius in a bootstrap variable, it will take effect with rounded-end, no?

cschweipert: It did and I tried to use round-end but it must get overwritten by something.

cschweipert: I tried to keep rounded-end instead but it must get overwritten somehow.

ethn (approved):
I might play with that at some point to understand, but it's a tiny issue in any case. I'm going to approve this now and get it deployed. Great work!

PR #1552, opened 2023-10-03
Testing attribution

cschweipert:


cschweipert:
I can't see the details why the tests failed on semaphore because I don't have access to the account.

ethn:
Looks like you have unintentional changes to `vendor/card-mods`.  I strongly suspect that's why the tests are failing.

PR #1554, opened 2023-10-06
Testing attribution

cschweipert:


ethn (changes requested):
I'd like to see the following tested in rspec to make sure we like the responses:

1. attributions view for one dataset (make sure all the tab content, box, etc is there)
2. attribute view for 
 - a dataset with an alteration
 - a metric from a community-assessed metric (make sure community credit appears)
 - a metric from a designer-assessed metric (make sure it doesn't)
 - an answer
  
You can pick one of those and do it in plain text format rather than html.

cschweipert:
@ethn I think the remaining codeclimate issues need to be ignored.

mod/deckorate_attribution/spec/attribution_spec.rb
```diff
@@ -0,0 +1,62 @@
+RSpec.describe Card::Set::Type::Reference do
```
ethn: do you understand the error message above? It is correct that the main "describe" line should line up with the relevant code. If this is code to test `Card::Set::Type::Reference`, the spec should be at the location mentioned in the error.

mod/deckorate_attribution/spec/attribution_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  def card_subject_name
+    "Reference 0000001"
+  end
+  check_views_for_errors
+
+  context "when a dataset with an alteration." do
+    let(:card) { Card.fetch("Reference 0000001") }
+
+    it "shows who did the adaptation in Rich Text." do
+      expect(card.format.render(:rich_text_attrib)).to match(/adaptation by Joe User/)
```
ethn: This is synonymous with:
` expect_view(:rich_text_attrib).to match(/adaptation by Joe User`

mod/deckorate_attribution/spec/attribution_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+    it "shows who did the adaptation in Plain Text." do
+      expect(card.format.render(:plain_text_attrib)).to match(/adaptation by Joe User/)
+    end
+    it "shows who did the adaptation in HTML." do
+      expect(card.format.render(:html_attrib)).to match(/adaptation by Joe User/)
+    end
+  end
+
+  context "when a metric from a community-assessed metric." do
```
ethn: perhaps it will read more naturally as "when metric is designer assessed"

ethn (changes requested):
Have a look at the code climate errors. We don't necessarily have to fix the "similar blocks of code" thing, but it's possible that using `expect_view` will fix that (because there is less code and therefore less code duplication that way).

We do definitely have to get the tests running.  You will likely need to run `bundle update` for that to happen. The issue is that the versions were updated in several of the gemspecs in card-mods, and therefore the current Gemfile.lock doesn't match the current gems, and therefore `bundle install` is failing on semaphore (where we're using `deployment mode, which gets grumpy when the Gemfile and Gemfile.lock are not lining up)

ethn (changes requested):
My guess is that the tests that are failing are doing so because of the changes to submodules.

Note that running `git submodule update` won't help you here. At least not now. The way that git submodules work is that a commit id is stored that indicates which version of the submodule you should use.  The fact that changes are showing up here means that you committed whatever version you had in your local copy of the repo. If you had run `git submodule update` before committing, it would have updated your local code based on the commit id.  Instead, you commited code (perhaps a commit all) and overrode the commit id. So now to fix it, you will have to get the right commit id back.

There are several ways to do this. One is to do a checkout of the code based on a branch that has the right submodule version. Eg `git checkout main vendor/decko` (assuming main has the right commit id.)

Another is to cd into vendor/decko and then checkout the main branch (making sure your local main branch is up to date.

In any case, please make sure that whenever you pull new code, AND DEFINITELY before you commit all, that you check to make sure your submodules are up to date.

ethn (changes requested):
Three specs are not passing on semaphore, because the new reference in the test yml is not in the fixtures. `tbe rake card:seed:update`

mod/deckorate_attribution/cypress/generate-attribution.spec.coffee
```diff
@@ -0,0 +1,53 @@
+describe "Attribution Generator - Dataset", ->
+  before ->
+    cy.login()
+    cy.visit("http://localhost:5002/Evil_Dataset")
```
ethn: We don't want/need the full localhost url here

ethn: In fact, we really don't want any references to localhost in this file

mod/deckorate_attribution/cypress/generate_attribution.spec.coffee
```diff
@@ -0,0 +1,53 @@
...
+    cy.get('button.submit-button').click();
+  it "generates attribution for Rich Text.", ->
+    cy.get(".tab-li-rich_text").click()
+    cy.contains("Wikirate.org, 'Evil Dataset' by Decko Bot, licensed under CC BY 4.0")
+  it "generates attribution for Plain Text.", ->
+    cy.get(".tab-li-plain_text").click()
+    cy.contains("Wikirate.org, 'Evil Dataset' by Decko Bot, licensed under CC BY 4.0 (https://creativecommons.org/licenses/by/4.0)")
+  it "generates attribution for HTML.", ->
+    cy.get(".tab-li-html").click()
+    cy.contains('<a href="https://wikirate.org" target="_blank">Wikirate.org</a>, \'<a href="http://localhost:5002/~3644" target="_blank">Evil Dataset</a>\' by <a href="http://localhost:5002/~1" target="_blank">Decko Bot</a>, licensed under <a href="https://creativecommons.org/licenses/by/4.0" target="_blank">CC BY 4.0</a>')
```
ethn, line 20: We definitely don't want card ids in here (3644, 1, etc). These ids (well, not the 1, but the 3644) could change every time we update the seed data

ethn (changes requested):

PR #1556, opened 2023-10-16
Drafting demo for odyssey

vasgat:


mod/wikirate_companies/set/type/wikirate_company/export.rb
```diff
@@ -10,6 +10,66 @@
...
+
+  def format_relationships answers
+    items = []
+    answers.each do |answer|
+      items.append({ metric: answer.metric.name,
+                     subject_company: odyssey_url(answer.subject_company_id),
+                     object_company: odyssey_url(answer.object_company_id),
+                     value: answer.value,
+                     year: answer.year,
+                     source: answer.source.card&.name
```
ethn, line 49: since every value in the hash is on the answer object, I would be tempted to put this code on the answer object.

ethn: (I mean relationship answer in this case. In the other case it's a normal answer)

ethn (approved):
I had a small suggestion, but we can merge this as a spike

ethn:
please make PR to main rather than staging so we can see all tests

PR #1562, opened 2023-10-31
modify cypress tests

cschweipert:


mod/deckorate_attribution/cypress/generate_attribution.spec.coffee
```diff
@@ -17,12 +17,12 @@ describe "Attribution Generator - Dataset", ->
     cy.contains("Wikirate.org, 'Evil Dataset' by Decko Bot, licensed under CC BY 4.0 (https://creativecommons.org/licenses/by/4.0)")
```
ethn: This should be angle braces. Can you change parenthesis to angle braces in the `:att_license` view and then fix the test to match?

mod/deckorate_attribution/cypress/generate_attribution.spec.coffee
```diff
@@ -17,12 +17,12 @@ describe "Attribution Generator - Dataset", ->
...
   it "generates attribution for HTML.", ->
     cy.get(".tab-li-html").click()
-    cy.contains('<a href="https://wikirate.org" target="_blank">Wikirate.org</a>, \'<a href="http://localhost:5002/~3644" target="_blank">Evil Dataset</a>\' by <a href="http://localhost:5002/~1" target="_blank">Decko Bot</a>, licensed under <a href="https://creativecommons.org/licenses/by/4.0" target="_blank">CC BY 4.0</a>')
+    cy.contains("div", /<a\s+href="((?!localhost|ids).+?)"\s+target="_blank">(.+?)<\/a>/)
```
ethn, line 20: This is fine. It seems a little funny that we're confirming that the localhost (which isn't part of the spec) is there but not confirming that "Evil Dataset" is, which is. But that's pretty thoroughly covered by the previous tests.

In general, the detailed testing should be in rspecs; in cypress we want to test just enough to confirm that the process is working. (And we don't want to have to go back and tweak the cypress tests every time, for example, parentheses turn to angle braces)

mod/deckorate_attribution/cypress/generate_attribution.spec.coffee
ethn: (This is a fix of the code, not just the cypress tests, but it should be quick)

ethn (changes requested):
just need the angle brackets fix (not a problem you introduced!)

PR #1568, opened 2023-11-22
Breadcrumbs

cschweipert:
Adds the link to the parent to breadcrumps if the breadcrumb has a parent and if the current card is the breadcrumb with the parent.


cschweipert:
This is work in progress. The tests revealed a bug.

mod/deckorate_datasets/set/type/dataset.rb
```diff
@@ -9,14 +10,38 @@
...
+    ]
+
+    insert_parent_link(breadcrumb_array) if should_insert_parent_link?
+
+    breadcrumb_array
+  end
+
+  private
+
+  def should_insert_parent_link?
```
ethn: more idiomatic just to say `#insert_parent_link?`

mod/deckorate_datasets/set/type/dataset.rb
```diff
@@ -9,14 +10,38 @@
...
+    card.parent != "" && card
```
ethn: I don't think card will ever be nil.

You could just do `card.parent.present?`

mod/deckorate_datasets/set/type/dataset.rb
```diff
@@ -9,14 +10,38 @@
...
 
-def parent_dataset_card
-  Card[parent_dataset]
+format :html do
+  def breadcrumb_items
+    breadcrumb_array = [
+      link_to("Home", href: "/"),
+      link_to_card(:dataset, "Datasets"),
+      render_name
+    ]
```
ethn: you could just say `breadcrumb_array = super`, right? Less duplication, more robust if the original method is edited?

mod/deckorate_datasets/set/type/dataset.rb
```diff
@@ -1,6 +1,7 @@
 include_set Abstract::Thumbnail
 include_set Abstract::DeckorateTabbed
 include_set Abstract::Bookmarkable
+include_set Abstract::Breadcrumbs
```
ethn: `Abstract::DeckorateTabbed` includes `Abstract::Header`, which includes `Abstract::Breadcrumbs`. So this is fine but not really necessary.

ethn (changes requested):
getting there! Just some very minor tweaks suggested

ethn (approved):
nice

PR #1570, opened 2023-11-29
Validate oc keys

cschweipert:


mod/wikirate_companies/set/type_plus_right/region/oc_jurisdiction_key.rb
```diff
@@ -5,3 +5,11 @@ def oc_code
 event :clear_jurisdiction_key_cache do
   Card::Region.cache.reset_all
 end
+
+event :validate_jurisdiction_key, :validate,
+      on: :save, changed: :content do
+  oc_key = Card.search left: { type: :region },
+                       right: :oc_jurisdiction_key,
+                       content: card.content
+  errors.add :content, "OC key already exists" if oc_key.present?
```
ethn: Let's write out "Open Corporates".

mod/wikirate_companies/set/type_plus_right/region/oc_jurisdiction_key.rb
```diff
@@ -5,3 +5,11 @@ def oc_code
 event :clear_jurisdiction_key_cache do
   Card::Region.cache.reset_all
 end
+
+event :validate_jurisdiction_key, :validate,
+      on: :save, changed: :content do
+  oc_key = Card.search left: { type: :region },
+                       right: :oc_jurisdiction_key,
+                       content: card.content
```
ethn, line 13: Technically this search will return an array (a list of cards). So even though it doesn't matter, it's slightly clearer to say `oc_keys`.

ethn (approved):
This looks very good. Thank you!

mod/wikirate_companies/set/type_plus_right/region/oc_jurisdiction_key.rb
cschweipert: Technically there should be only one element in the array. If I change it, oc_keys.present seems slightly less clear.

PR #1575, opened 2023-12-04
add country code validation

cschweipert:


cschweipert:
@ethn Do you think we can disregard the code climate issues (code duplication) because it's the nature of the rspec tests?

ethn (approved):
I do think we should have better helper methods for uniqueness verification. But for now I agree, we can let the duplication issues slide.

PR #1578, opened 2023-12-08
UI improvements

cschweipert:


mod/wikirate/assets/style/about_pages.scss
```diff
@@ -357,3 +352,16 @@
     }
   }
 }
+
+.slick-current {
+  .nodblclick {
+      background-color: #6B5CD8;
+      color: #fff;
```
ethn: can you use `$white` instead of `#fff`?

Currently those are the same, but if we use variables, then someone can change it to a different variant of the color and it will take effect everywhere.

mod/wikirate/set/self/homepage/involved.haml
```diff
@@ -4,9 +4,10 @@
     
   .py-5
     .homepage-involved-cards{ data: {
-      slick: { infinite: false,
+      slick: { infinite: true,
```
ethn: does the infinite make it go in a continuous loop? Was that requested? I thought before she said she didn't want that.

ethn: What story does this address?

ethn (changes requested):
tiny change plus a couple of questions

mod/wikirate/set/self/homepage/involved.haml
cschweipert: @ethn It addresses task two of this story #184726596. Without infinite = true the focused (purple) slide stops at the 3rd to last slide. Do you have different solution?

mod/wikirate/assets/style/about_pages.scss
```diff
@@ -357,3 +354,16 @@
     }
   }
 }
+
+.slick-current {
+  .nodblclick {
+      background-color: #6B5CD8;
+      color: $white;
+  }
```
ethn: `.slick-current` is applied to any item using the slick library. I don't think we really want to attach the styling to that.

`.nodblclick` is a functional class not intended for styling. (It should probably be preceded by an underscore; I will make a note to fix that). The thing we're styling here is the action-card.  Perhaps `.action-card > card` is more precise?  In fact, perhaps all of this should be inside `.slick-current .action-card`?

mod/wikirate/set/self/homepage/involved.haml
ethn: I'm struggling to connect this "3rd to last slide" comment to the story. Sorry if I'm being thick; it's a long story with lots of comments. Maybe can you show me today?

My concern is that maybe we're trying to address one ux design request by changing other parts of the design. The behavior was specified NOT to be infinite when we implemented it, so we can't just decide to change that without UX (melanie) approval.

ethn (changes requested):
I want to improve the selectors on the "current" action box styling.

Also want to understand the infinite issue, but I think I will need to see it. Can you show me during our call today?

ethn (approved):

PR #1581, opened 2023-12-22
Error message

cschweipert:


ethn (changes requested):
Thanks! Can you address the code climate concerns and use the locales file to internationalize the message? Thank you!

mod/deckorate_core/set/type/metric/answer_validation.rb
```diff
@@ -16,7 +16,8 @@ def validate_all_numeric_values
   bad_answer = metric_card.answers.find do |answer|
     !valid_numeric_value? answer.value
   end
-  "Non-numeric value: '#{bad_answer.value}'" if bad_answer.present?
+  return t(:answer_validation_error_message, bad_answer: bad_answer.value) \
+    if bad_answer.present?
```
ethn: CodeClimate didn't object to this? The `return` is superfluous, and the backslash makes things hard to read.

more typical would be:

```
  ...
  return unless bad_answer.present?

  t :answer_validation_error_message, bad_answer: bad_answer.value
end
```

ethn (changes requested):

PR #1584, opened 2024-01-16
Wikipedia companies

cschweipert:
Creates an official company website card. Disregard if this is not the route we want to take.


mod/deckorate_graphql/lib/graph_q_l/types/company.rb
```diff
@@ -4,6 +4,7 @@ module Types
     class Company < DeckorateCard
       field :headquarters, String, null: true
       field :wikipedia, String, null: true
+      field :official_company_website, String, null: true
```
vasgat, line 7: @ethn can we instead use the :wikirate_website card

ethn: yes!

ethn:
Sounds like this solution was rejected in favor of using the other codename, right?

PR #1602, opened 2024-04-13
Add a Dockerfile for deployments

AndrewKostka:


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    apt-get install -y nodejs
+
+COPY . .
+
+ENV RAILS_ENV=production
+
+RUN cp -R config/sample/* config && \
+    rm -R config/sample
+
+RUN bundle config without test cucumber cypress development profile
```
ethn, line 18: alternatively you could copy a .bundle/config file. Not a dramatic difference, but maybe some day we'll want to tweak other bundle configurations...

Gemfile
```diff
@@ -96,7 +97,6 @@ gem "timecop", group: %i[test cucumber] # date/time manipulation in tests
 group :test, :development do
   # gem "debase"
   gem "decko-spring", path: "./vendor/decko/support"
-  gem "puma", "~>5.6"                   # local webserver (6.x broke semaphore )
```
ethn, line 99: Is the notion that this is an interim solution and that we'll move to something more powerful before going live? Puma is a pretty lightweight webserver. I'd have to look a little more closely at our apache configurations to make specific arguments about what it can and can't do. 

But maybe the idea is that since each container has its own webserver, ligthweight is best?

ethn (approved):
Some minor comments/questions, but this looks very solid and mergeable to me.

btw, I would recommend that you make PRs to main branch. That triggers more thorough testing.

PR #1606, opened 2024-05-05
Automate Docker builds using Semaphore

AndrewKostka:


ethn (approved):
This looks really good.

Is is safe to assume a docker build failure would register as an error in semaphore?

PR #1701, opened 2024-10-26
Replace Puma with Passenger in Docker container

AndrewKostka:


docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM phusion/passenger-full:3.0.7
+
+RUN bash -lc "rvm --default use ruby-3.3.4"
+
+RUN apt-get update && \
```
ethn, line 5: I was thinking about writing in some short comments about the dependencies for folks skimming this. Is there any reason (convention or otherwise) to favor `&&` vs separate `RUN` statements?  (I was thinking for comments separate statements might be cleaner...

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    apt-get install -y imagemagick libcurl4-openssl-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+RUN curl -L -O https://github.com/a8m/envsubst/releases/download/v1.4.2/envsubst-Linux-x86_64 && \
+    echo "19b99358ec5db9205209072febdcbf5a  envsubst-Linux-x86_64" | md5sum -c - && \
+    chmod +x envsubst-Linux-x86_64 && \
+    mv envsubst-Linux-x86_64 /usr/local/bin/envsubst
+
+RUN rm -f /etc/service/nginx/down && \
```
ethn, line 15: Do we have to do this for the environment substitutions to work?

docker/nginx.conf
```diff
@@ -0,0 +1,16 @@
+server {
+  listen 80;
+
+  root /home/app/wikirate/public;
+
+  passenger_enabled on;
+  passenger_user app;
+
+  passenger_ruby /usr/bin/ruby3.3;
```
ethn, line 9: I'll switch semaphore testing to ruby 3.3. They have 3.3.0. Probably close enough.

ethn (approved):
This all looks great. A few questions in comments (just curiosity).

docker/Dockerfile
AndrewKostka: It's mainly to reduce the number of layers that are being created, but it can be a fine balance between that and readability. This post does a good job at explaining it: https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-which-is-better/39330458#39330458

docker/Dockerfile
AndrewKostka: Unfortunately, the built-in `envsubst` does not support default values.

PR #1839, opened 2025-08-12
Jsonld

vasgat:


mod/deckorate_core/set/type/answer/jsonld.rb
```diff
@@ -0,0 +1,47 @@
+format :jsonld do
+
+    def atom
```
ethn: 4 spaces. should be 2

mod/deckorate_core/set/type/answer/jsonld.rb
```diff
@@ -0,0 +1,47 @@
...
+            "license" => license_url(metric)
+        }.compact
+    end
+
+    def license_url metric
+        dir = metric.license.gsub(/(CC|4.0)/, "").strip.downcase
+        "https://creativecommons.org/licenses/#{dir}/4.0/"
+    end
+
+    def get_value(metric)
```
ethn: Can you add a `TODO` for me to find a more general place to put this? (Not your problem)

mod/deckorate_core/set/type/answer/jsonld.rb
```diff
@@ -0,0 +1,47 @@
...
+            "unit" => get_unit(metric),
+            "year" => card.year,
+            "source" => get_sources, 
+            "license" => license_url(metric)
+        }.compact
+    end
+
+    def license_url metric
+        dir = metric.license.gsub(/(CC|4.0)/, "").strip.downcase
+        "https://creativecommons.org/licenses/#{dir}/4.0/"
```
ethn: this is duplicate functionality.  It belongs in right/license.rb in the deckorate_utility mod. you could eg make a `:url` view in the base format and then do `field_nest :license, view :url`. The logic is already there.

mod/deckorate_core/set/type/answer/jsonld.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def get_value(metric)
+        metric.value_type == "Multi-Category" ? card.value&.split(", ") : card.value
+    end
+
+    def get_sources
+        sources = card.source&.split("\n") || []
```
ethn: this is cleaner: `sources = card.source_card.item_names`

ethn: ...then `return nil unless sources.present?` (no ampersand needed; present works with nil)

mod/deckorate_core/set/type/answer/jsonld.rb
```diff
@@ -0,0 +1,47 @@
...
+        sources&.any? ? sources.map { |source| path(mark: source, format: nil) } : nil
+    end
+
+    def get_unit metric
```
ethn: yeah, another TODO here would be nice. this is not jsonld-specific logic.

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -108,3 +108,65 @@ def semicolon_separated_values column
...
+  
+  def molecule
+    metric_jsonld(atom)
+  end
+
+  private
+
+    def metric_jsonld(a)
+    {
+      "@context" => "#{request.base_url}/context/#{card.type}.jsonld",
```
ethn: You can make a method for the context url in an abstract set and reuse it. It's the same everywhere

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -108,3 +108,65 @@ def semicolon_separated_values column
...
+      "@id"      => path(mark: card.name, format: nil),
+      "@type"    => card.type,
+
+      "title"    => a[:title] || a["title"],
```
ethn: if you call `symbolize_keys!` on the hash, then all the keys will be symbols, and you won't have to do the `symbol || string` logic over and over again

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -108,3 +108,65 @@ def semicolon_separated_values column
...
+      "value_options"               => get_elements(card.value_options),
+      "value_range"                 => a[:value_range] || a["value_range"],
+      "unit" => a[:unit] || a["unit"],
+      "formula" => get_formula,
+      "license" => license_url
+    }.compact
+  end
+
+  def license_url
+    dir = card.license.gsub(/(CC|4.0)/, "").strip.downcase
```
ethn: yeah, we don't want loads of copies of this logic...

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -108,3 +108,65 @@ def semicolon_separated_values column
...
+      "@type"    => card.type,
+
+      "title"    => a[:title] || a["title"],
+      "designer" => "#{request.base_url}/#{a[:designer] || a["designer"]}",
+      "question"  => a[:headquarters] || a["headquarters"],  
+
+      "metric_type"                 => a[:metric_type] || a["metric_type"],
+      "research_policy"             => a[:research_policy] || a["research_policy"],
+      "about"                       => ActionView::Base.full_sanitizer.sanitize(a[:about] || a["about"]),
+      "methodology"                 => ActionView::Base.full_sanitizer.sanitize(a[:methodology] || a["methodology"]),
```
ethn: a sanitize method in this class that calls the ActionView sanitizer would be cleaner

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -108,3 +108,65 @@ def semicolon_separated_values column
...
+    topics&.any? ? topics.map { |path| path(mark: path, format: nil) } : nil
+  end
+
+  def get_formula
+    formula = card.formula
+    formula.empty? ? nil : formula
+  end
+
+  def get_elements subcard
+    subcard&.any? ? subcard : nil
```
ethn: I think you should be able to handle all this nil stuff with a single method as long as you use the `present?` test (or the `blank?` test, which is its opposite. There's a lot of code here that's just handling nil, You could just use `presence` in a bunch of these cases, no?

mod/deckorate_research/set/type/relationship/jsonld.rb
```diff
@@ -0,0 +1,44 @@
...
+            "value" => get_value(metric),
+            "unit" => metric.unit.presence,
+            "year" => card.year,
+            "source" => get_sources, 
+            "license" => license_url(metric)
+
+        }.compact
+    end
+
+    def license_url metric
```
ethn: all the code below is duplicate, right?

You can either put it in a shared set or an abstract one. Eg Abstract::JsonldSupport...

mod/deckorate_research/set/type/source/jsonld.rb
```diff
@@ -0,0 +1,36 @@
...
+            "file" => card.file,
+            "report_type" => get_report_type, 
+            "company" => get_company,
+            "year" => get_year       
+
+        }.compact
+    end
+
+    def get_report_type
+        card.report_type&.split("\n").presence
```
ethn: `report_type_card.item_names`

There are several other instances of this...

mod/wikirate_companies/set/type/company/export.rb
```diff
@@ -7,20 +7,6 @@
     []
   end
 
-  view :odyssey do
```
ethn, line 10: farewell, sweet odyssey...

mod/topics/set/type/topic/export.rb
```diff
@@ -7,3 +7,36 @@ def atom
...
+    return nil unless category
+    path mark: category&.item_cards[0].name if category
+  end
+
+  def get_children
+    subtopics = card.fetch("subtopics")
+    
+    return [] unless subtopics
+
+    subtopics.item_cards.map do |c|
```
ethn: if you iterate over `item_names` rather than `item_cards`, you never have to instantiate the cards

mod/topics/set/type/topic/export.rb
```diff
@@ -7,3 +7,36 @@ def atom
...
+      "inScheme": path(mark: card.name.left), 
+      "parent": get_parent,
+      "children": get_children
+    }.compact
+  end
+
+  private
+
+  def get_parent
+    category = card.fetch(:category)
```
ethn: `category = card.category_card`

(Note, this works because of `card_accessor :category, type: :pointer` in topic.rb)

mod/wikirate_companies/set/type/company/export.rb
```diff
@@ -130,3 +72,58 @@ def identifier_cards names
...
+
+  private
+
+  def company_jsonld(a)
+    {
+      "@context" => "#{request.base_url}/context/#{card.type}.jsonld",
+      "@id"      => path(mark: card.name, format: nil),
+      "@type"    => card.type,
+
+      "name"     => a[:name] || a["name"],
```
ethn: see the `symbolize_keys!` comment above

mod/wikirate_companies/set/type/company/export.rb
```diff
@@ -130,3 +72,58 @@ def identifier_cards names
...
+      "alias"    => a[:alias] || a["alias"],
+      "logo"     => a[:image] || a["image"],
+      "country"  => get_country,
+
+      # identifiers (map 1:1 from atom)
+      "lei"                         => a[:legal_entity_identifier] || a["legal_entity_identifier"],
```
ethn: I don't like this. the code isn't supposed to know anything about  most of these company identifiers; they're all data with no codename. Everything will break if we change a cardname.

I think the solution should be to make it so that you can specify in each company identifier card (a) whether to include it in jsonld and (b) how to create the links. If this is too much for now, we can make a ticket it for it, but we shouldn't put it off too long.

public/context/Answer.jsonld
```diff
@@ -0,0 +1,17 @@
+{
```
ethn, line 1: is it idiomatic to capitalize these files? If not we can fix it.

ethn (changes requested):
Lots of great work here. I had a few suggestions...

mod/deckorate_core/set/type/answer/jsonld.rb
vasgat: I will add this TODO on the helper on the decko repo, I moved this repeated code on the JsonldSupport but I am not sure if it belongs there since it is quite wikirate specific work.

mod/wikirate_companies/set/type/company/export.rb
vasgat: I added a TODO here

public/context/Answer.jsonld
vasgat: will have a closer look on this

PR #1859, opened 2025-09-15
restrict exports to signed-in users

ethn:


ethn:
We decided not to move forward with this restriction.

PR #1873, opened 2025-10-22
Improve JSON-LD exports: add tests and fix metric/base_url handling

vasgat:


mod/deckorate_research/spec/set/type/relationship/jsonld_spec.rb
```diff
@@ -0,0 +1,14 @@
+RSpec.describe Card::Set::Type::Relationship::Jsonld do
+  let(:name) { "Commons+Supplied by+Monster_Inc+1977+Los_Pollos_Hermanos" }
```
ethn, line 2: is there a reason for separating out this name variable?

You could just do 

```
def card_subject
  "Commons+Supplied by+Monster_Inc+1977+Los_Pollos_Hermanos" .card
end
```

ethn (approved):
Looks great!

PR #1878, opened 2025-11-19
Jsonld

vasgat:
**Summary**

This PR updates the ontology and API to properly support framework mappings for metrics and improves the JSON/JSON-LD exports for both metrics and topics.

**Key Changes**

Ontology: Added wr:frameworkMapping (Metric → Topic) and aligned the JSON-LD context with this new property.

Metric Answers (JSON): Added unit, reorganized fields, and removed blank values for cleaner output.

Metric Export (JSON/JSON-LD): Exposed framework_mappings using topic_frameworks, ensured consistent units, and compacted empty fields.

Topic Export (JSON/JSON-LD): Added parent and children, switched metrics/datasets to API paths, removed bookmarkers, and cleaned blank values.

These changes improve clarity, consistency, and alignment between the ontology, API, and JSON-LD representations.


mod/deckorate_core/set/type/answer/json.rb
```diff
@@ -4,16 +4,21 @@ def atom
...
+
+      atom.delete_if { |_k, v| v.blank? }
     end
   end
 
   def molecule
-    super.merge sources: field_nest(:source, view: :items),
-                checked_by: field_nest(:checked_by, view: :items)
+    super.merge(sources: field_nest(:source, view: :items),
+                checked_by: field_nest(:checked_by, view: :items)).compact_blank!
```
ethn: As written, the molecule has both `source` and `sources`. We should delete `source`, right?

mod/deckorate_core/set/type/answer/json.rb
```diff
@@ -27,4 +32,12 @@ def item_cards
   def atom_content?
     false
   end
+
+  def get_unit metric
```
ethn: doesn't type/answer already have this method? It's in `Abstract::JsonldSupport`, right?

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -64,11 +64,11 @@ def semicolon_separated_values column
   def atom
     hash = super().merge designer: card.metric_designer, title: card.metric_title
     add_fields_to_hash hash, :core
-    hash
+    hash.compact_blank!
```
ethn: the `!` is superfluous in most if not all of these. It doesn't really matter that you change the variable value; you're only concerned about what is returned, right?

mod/topics/set/type/topic/export.rb
```diff
@@ -1,13 +1,27 @@
...
+      parent: get_parent,
+      children: get_children,
+      metrics: path(mark: card.metric_card.name),
+      datasets: path(mark: card.dataset_card.name)
+    ).compact_blank!
+  end
+
+  private
+
+  def get_parent
```
ethn: for what it's worth, I would recommend putting this method on the card, not the format. Doesn't matter much, but slightly more organized.  (Then you call `card.parent`)

mod/topics/set/type/topic/export.rb
```diff
@@ -1,13 +1,27 @@
...
+    return unless (category = card.category_card)
+
+    category.first_name.presence || card.name.left
```
ethn: Super nitpicky: `name.left` is a String, `name.left_name` is a Card::Name. So probably use `left_name` so that the method always returns the same kind of object.

mod/topics/set/type/topic/export.rb
```diff
@@ -30,7 +44,7 @@ def molecule
   def get_parent
     return unless (category = card.category_card)
 
-    path mark: category.first_name
+    category.first_name.present? ? path(mark: category.first_name) : path(mark: card.name.left)
```
ethn: or:

```
return unless (parent = super)

path mark: parent
```

ethn (changes requested):
Great stuff! A few things to look at.

mod/deckorate_core/set/type/answer/json.rb
vasgat: Kinda is. I had to remove the specific format to be able to use it also inside the :json format. 
on the mod/deckorate_utility/set/abstract/json_support I had to change "format :jsonld do" to "format do" so I can use it also inside the :json format. I am not sure if this is a good practice. Let me know if there is a better way to address thi.

ethn: That's a great solution.

mod/deckorate_core/set/type/metric/export.rb
```diff
@@ -127,12 +127,12 @@ def metric_jsonld(a)
       "title" => a[:title],
       "designer" => "#{request&.base_url}/#{a[:designer]}",
       "question" => a[:headquarters],
-
       "metric_type" => a[:metric_type],
       "research_policy" => a[:research_policy],
       "about" => sanitize_html(a[:about]),
       "methodology" => sanitize_html(a[:methodology]),
       "topics" => get_topics(a[:topics]),
+      "framework_mappings" => get_topics(a[:topic_frameworks]),
```
ethn, line 135: This is causing errors in the tests. I think the `#get_topics` method is expecting a topic, not a framework

mod/wikirate_companies/set/type/company/export.rb
```diff
@@ -11,13 +11,16 @@ def atom
     super.tap do |h|
       h.delete :content
       add_fields_to_hash h, :core
+
+      h.delete_if { |_k, v| v.blank? }
```
ethn: These are causing errors in the tests that are expecting nil values.

mod/topics/set/type/topic.rb
```diff
@@ -33,6 +33,12 @@ def stewarded_card
   framework_card
 end
 
+def parent
+  return unless (category = card.category_card)
+
+  category.first_card.presence || card.left
```
ethn: I don't think you want `card` in either of these.  (`card` is `self`).

mod/deckorate_utility/set/abstract/jsonld_support.rb
```diff
@@ -1,4 +1,4 @@
-format :jsonld do
+format do
```
ethn, line 1: This mostly makes sense, but the `context` method is jsonld-specific, so I'd probably put that one in a `format :jsonld` block

ethn (changes requested):
Just a couple more tiny changes (mostly to get the tests passing)

mod/deckorate_utility/set/abstract/jsonld_support.rb
vasgat: I have removed the jsonld context because the get_unit method is also being used under the json context as well. Let  me know if there is a better way to address this.

mod/deckorate_core/set/type/answer/json.rb
vasgat: I renamed the field to sources so it is overriden in molecule view

mod/deckorate_core/set/type/metric/export.rb
vasgat: Can we have a look together on this one? I do not see where in tests this can cause issues?

mod/wikirate_companies/set/type/company/export.rb
vasgat: I wanted to avoid returning null fields, I removed the changes since they are causing issues. There is no need to address this since we are aiming for code freeze. I will try to revoke similar changes I performed under different cardtypes.

